PR #3, opened 2024-11-11
Add support for DynamoDB Local

normj:
*Description of changes:*
Adds a new `AddAWSDynamoDBLocal` extension method for adding the DynamoDB Local container image to the Aspire application. When the DynamoDB Local Aspire resource is added as a reference to projects the `AWS_ENDPOINT_URL_DYNAMODB` environment variable is set. DynamoDB service clients created in the application that rely on the SDK to resolve the region/endpoint will pick up the environment variable. The `AWS_ENDPOINT_URL_DYNAMODB` will take precedence over the `AWS_REGION` environment variable.

Tests and options class exposing all of the options available for DynamoDB local have been added. PR is in draft mode till I confirm the experience with the DynamoDB team.

Sample of what the user experience is like.
```csharp
// Add a DynamoDB Local instance
var localDynamoDB = builder.AddAWSDynamoDBLocal("DynamoDBLocal");

// Reference DynamoDB local in project
builder.AddProject<Projects.Frontend>("Frontend")
        .WithReference(localDynamoDB);
```

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -4,24 +4,21 @@
     <CentralPackageTransitivePinningEnabled>true</CentralPackageTransitivePinningEnabled>
     <AspireVersion>9.0.0</AspireVersion>
   </PropertyGroup>
-
   <ItemGroup>
     <PackageVersion Include="Aspire.Hosting" Version="$(AspireVersion)" />
     <PackageVersion Include="Aspire.Hosting.AppHost" Version="$(AspireVersion)" />
+    <PackageVersion Include="AWSSDK.DynamoDBv2" Version="3.7.402.11" />
     <PackageVersion Include="Microsoft.Extensions.ServiceDiscovery.Dns" Version="$(AspireVersion)" />
-
     <PackageVersion Include="Microsoft.Extensions.Http.Resilience" Version="$(AspireVersion)" />
-
     <!-- AWS SDK for .NET dependencies -->
-    <PackageVersion Include="AWSSDK.CloudFormation" Version="3.7.400.36" />
-    <PackageVersion Include="AWSSDK.SQS" Version="3.7.400.36" />
-    <PackageVersion Include="AWSSDK.SimpleNotificationService" Version="3.7.400.36" />
-    <PackageVersion Include="AWSSDK.Core" Version="3.7.400.36" />
+    <PackageVersion Include="AWSSDK.CloudFormation" Version="3.7.400.47" />
+    <PackageVersion Include="AWSSDK.SQS" Version="3.7.400.47" />
+    <PackageVersion Include="AWSSDK.SimpleNotificationService" Version="3.7.400.47" />
+    <PackageVersion Include="AWSSDK.Core" Version="3.7.400.47" />
     <PackageVersion Include="AWSSDK.Extensions.NETCore.Setup" Version="3.7.301" />
     <PackageVersion Include="AWS.Messaging" Version="0.9.2" />
     <!-- AWS CDK dependencies -->
-    <PackageVersion Include="Amazon.CDK.Lib" Version="2.160.0" />
-
+    <PackageVersion Include="Amazon.CDK.Lib" Version="2.166.0" />
     <!-- Open Telemetry -->
     <PackageVersion Include="OpenTelemetry.Exporter.OpenTelemetryProtocol" Version="1.9.0" />
     <PackageVersion Include="OpenTelemetry.Extensions.Hosting" Version="1.9.0" />
@@ -29,7 +26,6 @@
     <PackageVersion Include="OpenTelemetry.Instrumentation.AWS" Version="1.1.0-beta.6" />
     <PackageVersion Include="OpenTelemetry.Instrumentation.Http" Version="1.9.0" />
     <PackageVersion Include="OpenTelemetry.Instrumentation.Runtime" Version="1.9.0" />
-
     <!-- Test dependencies -->
     <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.11.1" />
     <PackageVersion Include="xunit" Version="2.9.2" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -7,14 +7,15 @@
   <ItemGroup>
     <PackageVersion Include="Aspire.Hosting" Version="$(AspireVersion)" />
     <PackageVersion Include="Aspire.Hosting.AppHost" Version="$(AspireVersion)" />
-    <PackageVersion Include="AWSSDK.DynamoDBv2" Version="3.7.402.11" />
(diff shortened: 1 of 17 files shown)


davidfowl:
Consider using the `RunAsContainer` pattern [we use here](https://github.com/dotnet/aspire/blob/9abbb30cabe217bd0d03ecd9793e7a7427a71675/src/Aspire.Hosting.Azure.Redis/AzureRedisExtensions.cs#L178) 

```
var redis = builder.AddAzureRedis("redis").RunAsContainer();
```

src/Aspire.Hosting.AWS/DynamoDB/DynamoDBLocalOptions.cs
```diff
@@ -0,0 +1,52 @@
...
+
+    /// <summary>
+    /// The container image to run for DynamoDB local. The default is aws-dynamodb-local/aws-dynamodb-local.
+    /// </summary>
+    public string Image { get; set; } = "aws-dynamodb-local/aws-dynamodb-local";
+
+    /// <summary>
+    /// The container image tag. The default is latest.
+    /// </summary>
+    public string Tag { get; set; } = "latest";
```
davidfowl, line 26: Are you sure you want to use latest? It'll make people's experience unstable by default.

normj: I'll ping the DynamoDB team to see what they are comfortable with. They don't maintain a floating stable and latest version. So the alternative is hardcoding a specific version requiring a new release of the package every time there is a new version.

normj:
> Consider using the `RunAsContainer` pattern [we use here](https://github.com/dotnet/aspire/blob/9abbb30cabe217bd0d03ecd9793e7a7427a71675/src/Aspire.Hosting.Azure.Redis/AzureRedisExtensions.cs#L178)
> 
> ```cs
> var redis = builder.AddAzureRedis("redis").RunAsContainer();
> ```

My Azure ignorance here but what does `AddAzureRedis("redis")` do without the the `RunAsContainer()`? I assume that would be an indication when doing deployment that an Azure redis should be provisioned. In this case you would never provision the service DynamoDB.

eerhardt:
> what does AddAzureRedis("redis") do without the the RunAsContainer()? I assume that would be an indication when doing deployment that an Azure redis should be provisioned.

If you just call `.AddAzureRedis("redis")`, it will provision an Azure Cache for Redis both at F5 time and during publish. When you call `.AddAzureRedis("redis").RunAsContainer()`, it will use the local container during F5 time and it will provision an Azure Redis service during publish.

> In this case you would never provision the service DynamoDB.

Not even at publish time? What happens then?

normj:
> > In this case you would never provision the service DynamoDB.
> 
> Not even at publish time? What happens then?

Within DynamoDB you would provision tables but you don't provision the service. And the tables would be provision via the CloudFormation integration. This is all about setting up the emulator and redirecting the SDK when attempting to access the DynamoDB service go to the local emulator. 

There is a user experience question on how to create the tables within DynamoDB local because CloudFormation isn't going to do anything with the local emulator. So table creation for DynamoDB local would be done differently then when using the real service. Right now I would do it with adding a "Data Loader" project to the Aspire resource that took care of making sure DynamoDB local had all the table creation and data seeding. What I also want to experiment with is extending the `AddAWSDynamoDBLocal` with some fluent methods to define the local emulation tables.

Directory.Packages.props
```diff
@@ -2,38 +2,35 @@
...
-
   <ItemGroup>
     <PackageVersion Include="Aspire.Hosting" Version="$(AspireVersion)" />
     <PackageVersion Include="Aspire.Hosting.AppHost" Version="$(AspireVersion)" />
+    <PackageVersion Include="Aspire.Hosting.Testing" Version="$(AspireVersion)" />	
     <PackageVersion Include="Microsoft.Extensions.ServiceDiscovery.Dns" Version="$(AspireVersion)" />
-
-    <PackageVersion Include="Microsoft.Extensions.Http.Resilience" Version="9.0.0-preview.9.24507.7" />
-
+    <PackageVersion Include="Microsoft.Extensions.Http.Resilience" Version="9.0.0" />
```
philasmar, line 12: nit: why not use $(AspireVersion) here as well.

src/Aspire.Hosting.AWS/DynamoDB/DynamoDBLocalOptions.cs
```diff
@@ -0,0 +1,59 @@
...
+
+    /// <summary>
+    /// If set to true DynamoDB local uses a single database file instead of separate files for each credential and Region.
+    /// </summary>
+    public bool SharedDb { get; set; }
+
+    /// <summary>
+    /// If set to true DynamoDB runs in memory instead of using a database file. DynamoDB local will run faster
+    /// using InMemory mode but all data will be lost when the container ends and the data stored in DynamoDB
+    /// local can ont exceed the available memory for the container.
```
philasmar, line 36: nit: typo in "can ont"

tests/Aspire.Hosting.AWS.Integ.Tests/PlaygroundE2ETests.cs
```diff
@@ -0,0 +1,71 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+//using Aspire.Hosting.Testing;
```
philasmar, line 3: nit: remove commented out statement

tests/Aspire.Hosting.AWS.Tests/DynamoDBLocalCommandLineArgumentTests.cs
philasmar, line 1: this file doesn't have the Amazon header

src/Aspire.Hosting.AWS/DynamoDB/DynamoDBLocalOptions.cs
normj: Fixed

tests/Aspire.Hosting.AWS.Integ.Tests/PlaygroundE2ETests.cs
normj: Fixed

Directory.Packages.props
normj: Fixed after rebasing from `dev`.

tests/Aspire.Hosting.AWS.Tests/DynamoDBLocalCommandLineArgumentTests.cs
normj: Fixed

philasmar (approved):

gcbeattyAWS (approved):

PR #14, opened 2025-01-22
Use Aspire 9's WaitFor feature for making projects wait for CloudFormation

normj:
*Description of changes:*
To be consistent with other Aspire integrations use the `WaitFor` mechanism added in Aspire 9 for forcing projects to wait till CloudFormation is ready. This is a switch from the homegrown solution of using our own `TaskCompletionSource` to force a block in the project resource.

The `TaskCompletionSource` is still used in other parts of the integration, specifically the public `GetValueAsync` API for a `StackOutputReference` so the code still exist but for the common case of having the project reference the stack the experience will be consistent with other integrations.

### DIFF
--- a/.autover/changes/e537d29a-c69e-46f0-a857-68c6a75b23d6.json
+++ b/.autover/changes/e537d29a-c69e-46f0-a857-68c6a75b23d6.json
@@ -1,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "Aspire.Hosting.AWS",
+      "Type": "Patch",
+      "ChangelogMessages": [
+        "Use Aspire 9\u0027s WaitFor mechanism for waiting CloudFormation resource to be running"
+      ]
+    }
+  ]
+}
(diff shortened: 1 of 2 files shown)


gcbeattyAWS (approved):

.autover/changes/e537d29a-c69e-46f0-a857-68c6a75b23d6.json
```diff
@@ -0,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "Aspire.Hosting.AWS",
+      "Type": "Patch",
+      "ChangelogMessages": [
+        "Use Aspire 9\u0027s WaitFor mechanism for waiting CloudFormation resource to be running"
```
philasmar, line 7: nit: you might want to remove the unicode char since that's how it'll look in the changelog

philasmar (approved):

.autover/changes/e537d29a-c69e-46f0-a857-68c6a75b23d6.json
normj: Removed the Unicode character

Issue #17, opened 2025-01-28
Aspire Lambda Local Development Feature Tracker


normj:
The next major effort being worked on by the AWS team is build a local development environment for .NET Lambda functions. The work is being done across this aspire repository and the [aws/aws-lambda-dotnet](https://github.com/aws/aws-lambda-dotnet) where a new version of the Mock Lambda Test tool is being created to act as a Lambda service emulator. Work on the new version of the test tool is being done in the [feature/lambdatesttool-v2](https://github.com/aws/aws-lambda-dotnet/tree/feature/lambdatesttool-v2/Tools/LambdaTestTool-v2).

The plan for the initial scope of the Lambda integration with Aspire is to allow users to declare in the Aspire AppHost the Lambda functions. In addition add an API Gateway emulator attaching Lambda functions to the emulator.
```
using Aspire.Hosting.AWS.Lambda;

#pragma warning disable CA2252 // This API requires opting into preview features

var builder = DistributedApplication.CreateBuilder(args);

builder.AddAWSLambdaFunction<Projects.ToUpperLambdaFunctionExecutable>("ToUpperFunction", lambdaHandler: "ToUpperLambdaFunctionExecutable");

var defaultRouteLambda = builder.AddAWSLambdaFunction<Projects.WebDefaultLambdaFunction>("LambdaDefaultRoute", lambdaHandler: "WebDefaultLambdaFunction");
var addRouteLambda = builder.AddAWSLambdaFunction<Projects.WebAddLambdaFunction>("AddDefaultRoute", lambdaHandler: "WebAddLambdaFunction");
var minusRouteLambda = builder.AddAWSLambdaFunction<Projects.WebMinusLambdaFunction>("MinusDefaultRoute", lambdaHandler: "WebMinusLambdaFunction");


builder.AddAPIGatewayEmulator("APIGatewayEmulator", Aspire.Hosting.AWS.Lambda.APIGatewayType.HttpV2)
        .WithReference(defaultRouteLambda, Method.Get, "/")
        .WithReference(addRouteLambda, Method.Get, "/add/{x}/{y}")
        .WithReference(minusRouteLambda, Method.Get, "/minus/{x}/{y}");

builder.Build().Run();
```
When you run the AppHost you will be able to see the Lambda functions as resources and resources that represent emulators for Lambda and API Gateway.

![Image](https://github.com/user-attachments/assets/14764152-845c-4bb9-a6f4-244d94384bf8)

[...]

PR #27, opened 2025-02-04
Add OTEL support for Lambda projects in the Aspire Dashboard

normj:
*Issue #, if available:*
https://github.com/aws/integrations-on-dotnet-aspire-for-aws/issues/17#issuecomment-2628827085

*Description of changes:*
When adding the Lambda project resources the OpenTelemetry defaults that are normally set for a project in Aspire were not being done. That is because we don't go through the normal `AddProject` API because we need to return our own subclass. This PR adapts the project defaults by `AddProject` for what we need in a Lambda project. Which is basically not needing any of the Kestrel settings but taking all of the OpenTelemetry settings.

The Lambda projects in the playground were reworked to use an Aspire service defaults project and enable OpenTelemetry in each of the projects. I also added a new Lambda function to the playground that uses the AWS SDK to show telemetry from the AWS SDK was picked.

Navigating to the Lambda function's via the API Gateway emulator now shows the Lambda functions in the trace window of the Aspire Dashboard.

image

Drilling into the Lambda function that makes AWS calls shows this.

image


GarrettBeatty (approved):

src/Aspire.Hosting.AWS/Lambda/LambdaExtensions.cs
```diff
@@ -97,6 +96,33 @@ public static class LambdaExtensions
...
+    /// https://github.com/dotnet/aspire/blob/157f312e39300912b37a14f59beda217c8195e14/src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs#L287
+    /// </summary>
+    /// <param name="builder"></param>
+    /// <returns></returns>
+    private static IResourceBuilder<LambdaProjectResource> WithOpenTelemetry(this IResourceBuilder<LambdaProjectResource> builder)
+    {
+        builder.WithEnvironment("OTEL_DOTNET_EXPERIMENTAL_OTLP_EMIT_EXCEPTION_LOG_ATTRIBUTES", "true");
+        builder.WithEnvironment("OTEL_DOTNET_EXPERIMENTAL_OTLP_EMIT_EVENT_LOG_ATTRIBUTES", "true");
+        // .NET SDK has experimental support for retries. Enable with env var.
+        // https://github.com/open-telemetry/opentelemetry-dotnet/pull/5495
```
dscpinheiro, line 110: This PR is merged, is the support for retries still experimental?

dscpinheiro: Oh, I see that's the PR for the environment variable 🤦

It's still experimental according to this: https://github.com/vishweshbankwar/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/README.md#experimental-features

dscpinheiro (approved):

PR #28, opened 2025-02-06
Automatically install Amazon.Lambda.TestTool

normj:
*Description of changes:*
If the Lambda service emulator is added there will be a check made to see if Amazon.Lambda.TestTool is installed and up to date. If not then it will be installed via the `dotnet tool install` command.

The Lambda service emulator is added to the collection resources as part of the first invoke of `AddAWSLambdaFunction`. If the user wants to customize the installation or disable it they can call the new `AddAWSLambdaServiceEmulator` method before calling `AddAWSLambdaFunction`. The `AddAWSLambdaServiceEmulator` takes in an `LambdaEmulatorOptions` that allows disabling the auto install, overriding the version to install or allowing downgrading the version. By default if a newer version of the tooling is already installed then what the Aspire.Hosting.AWS is looking for we will allow assuming the user knows what they are doing.

While going through the area I did some minor clean work.
* Fix all the namespaces of the classes in the unit tests projects
* Made the pattern of adding the Aspire ExecutableResource that represents the emulators the same across Lambda and API Gateway. This including some class renames.

This PR is depending on the Amazon.Lambda.TestTool [[PR#](https://github.com/aws/aws-lambda-dotnet/pull/1969)](https://github.com/aws/aws-lambda-dotnet/pull/1969) that adds a new `--tool-info` switch that Aspire can used to get metadata about the currently installed version.


src/Aspire.Hosting.AWS/Utils/Internal/ProcessCommandService.cs
```diff
@@ -0,0 +1,101 @@
+﻿// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+using Microsoft.Extensions.Logging;
+using System.Diagnostics;
+using System.Text;
+
+namespace Aspire.Hosting.AWS.Utils.Internal;
```
normj, line 7: I didn't want to expose this interface as part of our public API but for unit testing purposes I need to implement the interface and making the interface internal doesn't work for that even marking the assembly internal visible to. So I went with the compromise of putting this in an "Internal" namespace.

src/Aspire.Hosting.AWS/Lambda/LambdaLifecycleHook.cs
```diff
@@ -0,0 +1,126 @@
...
+using Microsoft.Extensions.Logging;
+using System.Text.Json;
+using System.Text.Json.Nodes;
+using System.Threading;
+using static Google.Protobuf.Reflection.GeneratedCodeInfo.Types;
+
+namespace Aspire.Hosting.AWS.Lambda;
+
+/// <summary>
+/// Lambda lifecycle hook takes care of getting Amazon.Lambda.TestTool is installed if there was
```
GarrettBeatty, line 15: nit: `is installed` -> `installed`.

GarrettBeatty (approved):

src/Aspire.Hosting.AWS/Lambda/LambdaLifecycleHook.cs
normj: Fixed

Directory.Packages.props
philasmar, line 1: Unrelated to this change, but why does the entire solution has 1 Directory.Packages.props? So anytime we add a new package, we are adding it for the entire solution?

src/Aspire.Hosting.AWS/Lambda/APIGatewayEmulatorResource.cs
```diff
@@ -0,0 +1,26 @@
...
+/// </summary>
+/// <param name="name">Aspire resource name</param>
+public class APIGatewayEmulatorResource(string name, APIGatewayType apiGatewayType) : ExecutableResource(name, 
+        "dotnet",
+        Environment.CurrentDirectory
+        )
+{
+    internal void AddCommandLineArguments(IList<object> arguments)
+    {
+        arguments.Add("lambda-test-tool");
```
philasmar, line 20: isn't the name of our tool `dotnet-lambda-test-tool`?

src/Aspire.Hosting.AWS/Lambda/LambdaEmulatorAnnotation.cs
```diff
@@ -14,4 +14,22 @@ internal class LambdaEmulatorAnnotation(EndpointReference endpoint) : IResourceA
     /// The HTTP endpoint for the Lambda runtime emulator.
     /// </summary>
     public EndpointReference Endpoint { get; init; } = endpoint;
+
+    /// <summary>
+    /// If set to true Amazon.Lambda.TestTool will updated/installed during AppHost startup. Amazon.Lambda.TestTool is
```
philasmar, line 19: the name and the description are saying 2 different things. The name suggests that we won't auto install, but the description is saying that we will?

src/Aspire.Hosting.AWS/Lambda/LambdaEmulatorOptions.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+namespace Aspire.Hosting.AWS.Lambda;
+
+/// <summary>
+/// Options that can be added the Lambda emulator resource.
```
philasmar, line 6: nit: add "to" between "added" and "the"

src/Aspire.Hosting.AWS/Lambda/LambdaEmulatorOptions.cs
```diff
@@ -0,0 +1,27 @@
...
+/// </summary>
+public class LambdaEmulatorOptions
+{
+    /// <summary>
+    /// If set to true, Amazon.Lambda.TestTool will updated/installed during AppHost startup. Amazon.Lambda.TestTool is
```
philasmar, line 11: same comment as before

src/Aspire.Hosting.AWS/Lambda/LambdaEmulatorResource.cs
```diff
@@ -0,0 +1,22 @@
...
+/// </summary>
+/// <param name="name">Aspire resource name</param>
+public class LambdaEmulatorResource(string name) : ExecutableResource(name, 
+        "dotnet",
+        Environment.CurrentDirectory
+        )
+{
+    internal void AddCommandLineArguments(IList<object> arguments)
+    {
+        arguments.Add("lambda-test-tool");
```
philasmar, line 19: same question on tool name

philasmar: also didn't we change the behavior so that you need to specify the port for the emulator to run? won't this error out as is?

src/Aspire.Hosting.AWS/Lambda/LambdaExtensions.cs
```diff
@@ -97,32 +97,62 @@ public static class LambdaExtensions
         return resource;
     }
 
-    private static ExecutableResource AddOrGetLambdaServiceEmulatorResource(IDistributedApplicationBuilder builder)
-    {        
-        if (builder.Resources.FirstOrDefault(x => x.TryGetAnnotationsOfType<LambdaEmulatorAnnotation>(out _)) is not ExecutableResource serviceEmulator)
+    /// <summary>
+    /// Add the Lambda service emulator resource. The AddAWSLambdaFunction method will automatically add the Lambda service emulator if it hasn't
```
philasmar, line 101: could you add doc references to "AddAWSLambdaFunction" and "LambdaEmulatorOptions"

src/Aspire.Hosting.AWS/Lambda/LambdaLifecycleHook.cs
```diff
@@ -0,0 +1,126 @@
...
+
+        var installedVersion = Version.Parse(currentInstalledVersion.Replace("-preview", string.Empty));
+        var expectedVersion = Version.Parse(expectedVersionStr.Replace("-preview", string.Empty));
+
+        return (installedVersion < expectedVersion) || (allowDowngrading && installedVersion != expectedVersion);
+    }
+
+    private async Task<string> GetCurrentInstalledVersionAsync(CancellationToken cancellationToken)
+    {
+        var results = await processCommandService.RunProcessAndCaptureOuputAsync(logger, "dotnet", "lambda-test-tool --tool-info", cancellationToken);
```
philasmar, line 101: might need to update based on other PR changes https://github.com/aws/aws-lambda-dotnet/pull/1969

src/Aspire.Hosting.AWS/Lambda/APIGatewayEmulatorResource.cs
GarrettBeatty: apparently ` dotnet lambda-test-tool` works and so does `dotnet-lambda-test-tool`.

Directory.Packages.props
normj: The repo was setup so that it had centralized package versioning. But in the individual project files you still specify the package you want to use. But in the project you don't set a version. For example in the project file of Aspire.Hosting.AWS we add the packages but don't set the version.
```
  <ItemGroup>
    <PackageReference Include="Aspire.Hosting" />
    <PackageReference Include="Amazon.CDK.Lib" />
    <PackageReference Include="AWSSDK.Core" />
    <PackageReference Include="AWSSDK.CloudFormation" />
  </ItemGroup>
```

This way you only have to update the version of a package in one place for all projects in a repository.

src/Aspire.Hosting.AWS/Lambda/APIGatewayEmulatorResource.cs
normj: Discussed in sync meeting and we are on the same page now.

src/Aspire.Hosting.AWS/Lambda/LambdaEmulatorAnnotation.cs
normj: Fixed

src/Aspire.Hosting.AWS/Lambda/LambdaEmulatorOptions.cs
normj: Fixed

src/Aspire.Hosting.AWS/Lambda/LambdaEmulatorOptions.cs
normj: Fixed

src/Aspire.Hosting.AWS/Lambda/LambdaExtensions.cs
normj: Done

src/Aspire.Hosting.AWS/Lambda/LambdaLifecycleHook.cs
normj: Done

philasmar (approved):

PR #30, opened 2025-02-18
Add validation for AWS credentials

normj:
*Description of changes:*
To help users with trouble shooting AWS configuration add a validation for the AWS credentials that could be potentially used by the SDK. The validation is done as a non blocking background task to avoid slowing down the startup process.

This could be potentially expanded in the future for the AppHost to resolve credentials with SSO login being handled from AppHost. For now though it would let you know if the login command done by PowerShell or CLI hasn't been done is expired.

### DIFF
--- a/src/Aspire.Hosting.AWS/SdkUtilities.cs
+++ b/src/Aspire.Hosting.AWS/SdkUtilities.cs
@@ -3,7 +3,11 @@
 using System.Reflection;
 using System.Text;
 using Amazon.Runtime;
+using Microsoft.Extensions.Logging;
+using Amazon.SecurityToken;
+using Amazon.SecurityToken.Model;
 using Aspire.Hosting.ApplicationModel;
+using Amazon.Runtime.CredentialManagement;
 
 namespace Aspire.Hosting.AWS;
 internal static class SdkUtilities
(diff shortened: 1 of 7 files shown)


boblodgett (approved):

src/Aspire.Hosting.AWS/SdkUtilities.cs
```diff
@@ -65,4 +75,84 @@ internal static void ApplySDKConfig(EnvironmentCallbackContext context, IAWSSDKC
...
+        var defaultProfile = !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("AWS_PROFILE")) ? Environment.GetEnvironmentVariable("AWS_PROFILE") : SharedCredentialsFile.DefaultProfileName;
+        if (profiles.FirstOrDefault(x => string.Equals(defaultProfile, x.Name)) == null)
+            return;
+
+        _ = ValidateSdkConfigAsync(logger, new AWSSDKConfig(), true);
+    }
+
+    internal static void BackgroundSDKConfigValidation(ILogger logger, IAWSSDKConfig config)
+    {
+        _ = ValidateSdkConfigAsync(logger, config, false);
```
peterrsongg, line 99: just wondering why we use the discard variable here. Is it simply to avoid the warning that is thrown that the variable isn't being used?

peterrsongg (approved):
approved with a general question

src/Aspire.Hosting.AWS/SdkUtilities.cs
normj: Yes, to avoid the warning.

GarrettBeatty (approved):

PR #31, opened 2025-02-18
feat: add support for class library lambda functions

philasmar:
*Issue #, if available:*
DOTNET-7958

*Description of changes:*
* Add support for adding Class Library Lambda Functions. We are currently using Reflection to get access to an Annotation which enables us to use a Launch Profile for the Class Library. When Aspire 9.1 comes out, that annotation will be made public and we can get rid of the reflection.


src/Aspire.Hosting.AWS/Lambda/LambdaLifecycleHook.cs
```diff
@@ -22,10 +21,51 @@ internal class LambdaLifecycleHook(ILogger<LambdaEmulatorResource> logger, IProc
 
     public async Task BeforeStartAsync(DistributedApplicationModel appModel, CancellationToken cancellationToken = default)
     {
+        var classLibraryProjectPaths = 
+            appModel.Resources
+                .OfType<LambdaProjectResource>()
+                .Where(x =>
+                {
+                    if (!x.TryGetLastAnnotation<LambdaFunctionAnnotation>(out var functionAnnotation))
```
GarrettBeatty, line 29: what is the significance of using the last annotation? i understand this whole thing is supposed to filter for class library project paths but wasnt clear on this one line

src/Aspire.Hosting.AWS/Lambda/LambdaLifecycleHook.cs
```diff
@@ -123,4 +163,72 @@ private async Task<string> GetCurrentInstalledVersionAsync(CancellationToken can
             return string.Empty;
         }
     }
+
+    internal async Task<string> GetCurrentInstallPathAsync(CancellationToken cancellationToken)
```
GarrettBeatty, line 167: GetCurrentInstallPathAsync doesnt seem like it should go in lambdaLifecycleHook class. wondering if we can put it somewhere else?

src/Aspire.Hosting.AWS/Lambda/LambdaLifecycleHook.cs
```diff
@@ -123,4 +163,72 @@ private async Task<string> GetCurrentInstalledVersionAsync(CancellationToken can
...
+            return installPath ?? string.Empty;
+        }
+        catch (JsonException ex)
+        {
+            logger.LogWarning(ex, "Error parsing install path information from Amazon.Lambda.TestTool: {versionInfo}", results.Output);
+            return string.Empty;
+        }
+    }
+
+    internal async Task<string> GetProjectAssemblyNameAsync(string projectPath, CancellationToken cancellationToken)
```
GarrettBeatty, line 195: same here and other similar functions

src/Aspire.Hosting.AWS/Utils/ProjectUtilities.cs
```diff
@@ -0,0 +1,123 @@
...
+using System.Text.Json.Nodes;
+
+namespace Aspire.Hosting.AWS.Utils;
+
+internal static class ProjectUtilities
+{
+    /// <summary>
+    /// Initializes the project's launch settings if necessary, and
+    /// ensures they are referencing the Amazon.Lambda.TestTool's location.
+    /// </summary>
```
GarrettBeatty, line 12: can you add docs for each param. and maybe a link/sample of the generated profile?

src/Aspire.Hosting.AWS/Utils/ProjectUtilities.cs
```diff
@@ -0,0 +1,123 @@
...
+    /// <returns>The launchSetting.json content</returns>
+    private static string GetLaunchSettings(string projectPath)
+    {
+        var parentDirectory = Path.GetDirectoryName(projectPath);
+        if (string.IsNullOrEmpty(parentDirectory))
+            throw new ArgumentException($"The project path '{projectPath}' is invalid. Unable to retrieve the '{Constants.LaunchSettingsFile}' file.");
+        var properties = Path.Combine(parentDirectory, "Properties");
+        if (!Directory.Exists(properties))
+        {
+            Directory.CreateDirectory(properties);
```
GarrettBeatty, line 99: so i understand in this case the launch settings doesnt exist. im not sure if we should be creating a directory though here. i assume you do this because when you end up writing the new launch settings you need the directory to exist right? maybe have the creation of the directory in there.

GarrettBeatty (changes requested):

GarrettBeatty:
i see some of the deleted files are web lambda functions. im assuming those were remade into different files is that correct?

src/Aspire.Hosting.AWS/Lambda/LambdaLifecycleHook.cs
philasmar: Also following the convention that Norm started. We already make these kinds of calls in `LambdaLifecycleHook`. Check the above funciton for more context. Also, this is where we have access to the IProcessCommandService. If you feel very strongly about it, I could look into refactoring it.

src/Aspire.Hosting.AWS/Lambda/LambdaLifecycleHook.cs
philasmar: same comment as above

src/Aspire.Hosting.AWS/Lambda/LambdaLifecycleHook.cs
GarrettBeatty: i dont know the code well enough but is there any case where doing last vs first would have a different result?

philasmar: No they're the same. We only have one annotation in there. If for any reason a new one was added, taking the last one makes more sense than the first one.

src/Aspire.Hosting.AWS/Lambda/LambdaLifecycleHook.cs
```diff
@@ -22,10 +21,51 @@ internal class LambdaLifecycleHook(ILogger<LambdaEmulatorResource> logger, IProc
...
+            foreach (var projectResource in classLibraryProjectPaths)
+            {
+                var projectMetadata = projectResource.Annotations
+                    .OfType<IProjectMetadata>()
+                    .First();
+                var lambdaFunctionAnnotation = projectResource.Annotations
+                    .OfType<LambdaFunctionAnnotation>()
+                    .First();
+            
+                var installPath = await GetCurrentInstallPathAsync(cancellationToken);
```
normj, line 53: Should you add a check if `installPath` is empty string and if so log an error and continue? Actually looking at the utility methods `targetFramework` and `assemblyName` should have the check as well.

src/Aspire.Hosting.AWS/Utils/ProjectUtilities.cs
```diff
@@ -0,0 +1,123 @@
...
+        string assemblyName, 
+        string projectPath, 
+        string runtimeSupportAssemblyPath, 
+        string targetFramework)
+    {
+        // Retrieve the current launch settings JSON from wherever it's stored.
+        string launchSettingsJson = GetLaunchSettings(projectPath);
+
+        // Parse the JSON into a mutable JsonNode (root is expected to be an object)
+        JsonNode? rootNode = JsonNode.Parse(launchSettingsJson);
```
normj, line 25: Can you catch JSON parse exceptions in case the launchsettings json file is malformed and if so log an error and return.

src/Aspire.Hosting.AWS/Utils/ProjectUtilities.cs
```diff
@@ -0,0 +1,123 @@
...
+        if (rootNode is not JsonObject root)
+        {
+            // If the parsed JSON isn’t an object, initialize a new one.
+            root = new JsonObject();
```
normj, line 29: To my point above. If the launchsettings is malformed we should log and return. I would not want to replace the content which could cause users to lose useful config. If we log it then can fix the syntax error and restart debugging.

src/Aspire.Hosting.AWS/Utils/ProjectUtilities.cs
```diff
@@ -0,0 +1,123 @@
...
+                ["executablePath"] = "dotnet"
+            };
+
+            profiles[launchSettingsNodeKey] = lambdaTester;
+        }
+
+        // Update properties that contain a path that is environment-specific
+        lambdaTester["commandLineArgs"] =
+            $"exec --depsfile ./{assemblyName}.deps.json --runtimeconfig ./{assemblyName}.runtimeconfig.json {SubstituteHomePath(runtimeSupportAssemblyPath)} {functionHandler}";
+        lambdaTester["workingDirectory"] = Path.Combine(".", "bin", "$(Configuration)", targetFramework);
```
normj, line 54: Somewhere there should be a check that we have version of RuntimeSupport in a targetFramework that the Lambda function is targeting. If we don't have a match we should log error and continue.

src/Aspire.Hosting.AWS/Utils/ProjectUtilities.cs
normj: I agree that it was surprising the `GetLaunchSettings` creates the directory. I think you can move that to the `SaveLaunchSettings` which is where I would expect the creation to exist.

normj (changes requested):

philasmar:
> i see some of the deleted files are web lambda functions. im assuming those were remade into different files is that correct?

Yes, Norm consolidated everything into the WebCalculator project

src/Aspire.Hosting.AWS/Utils/AspireUtilities.cs
```diff
@@ -0,0 +1,8 @@
+﻿// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+namespace Aspire.Hosting.AWS.Utils;
+
+public static class AspireUtilities
```
normj, line 5: This should be `internal`.

src/Aspire.Hosting.AWS/Utils/Internal/ProcessCommandService.cs
```diff
@@ -98,4 +109,62 @@ internal class ProcessCommandService : IProcessCommandService
...
+    
+    /// <summary>
+    /// Utility method for running a command on the commandline. It returns backs the exit code.
+    /// </summary>
+    /// <param name="logger"></param>
+    /// <param name="path"></param>
+    /// <param name="arguments"></param>
+    /// <param name="workingDirectory"></param>
+    /// <returns></returns>
+    public int RunProcess(ILogger logger, string path, string arguments, string workingDirectory)
```
normj, line 121: Was there a reason the existing `RunProcessAndCaptureOuputAsync` didn't work? Looks like they do basically the same thing except the original also returned back the output. If you used that one then LambdaLifecycleHook that calls this we could check the error code and if it was non-Zero log the output and since in that context we know it is fatal use `LogError`?

src/Aspire.Hosting.AWS/Utils/ProjectUtilities.cs
```diff
@@ -0,0 +1,184 @@
...
+        Directory.CreateDirectory(tempPath);
+
+        var projectContent = $@"
+<Project Sdk=""Microsoft.NET.Sdk"">
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include=""Amazon.Lambda.RuntimeSupport"" Version=""1.12.2"" />
```
normj, line 162: Can you move the version into the same as an internal constant in the `Constants.cs` file? For consistency I would appreciate if you move the `DefaultLambdaTestToolVersion` constant from `LambdaLifecycleHook` to their as well. That way as we make new versions of the test tool or runtime support we just have one spot we need to update.

normj (changes requested):

src/Aspire.Hosting.AWS/Utils/ProjectUtilities.cs
philasmar: Done

src/Aspire.Hosting.AWS/Utils/Internal/ProcessCommandService.cs
philasmar: `RunProcessAndCaptureOuputAsync` is an async function while the new one isn't. I tried using the existing one but there's a deadlock happening somewhere. I ran 5 runs and they all timed out after 2 hours.

GarrettBeatty (approved):

playground/Lambda/WebCalculatorFunctions/Properties/launchSettings.json
```diff
@@ -0,0 +1,34 @@
+{
+  "profiles": {
+    "Mock Lambda Test Tool": {
```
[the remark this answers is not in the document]
philasmar: I removed it.

src/Aspire.Hosting.AWS/Lambda/LambdaExtensions.cs
```diff
@@ -35,12 +40,35 @@ public static class LambdaExtensions
...
+                var constructor = launchProfileAnnotationsType!.GetConstructors()[0];
+                var instance = constructor.Invoke(new object[] { $"{Constants.LaunchSettingsNodePrefix}{name}" }) as IResourceAnnotation;
+
+                var project = new LambdaProjectResource(name);
+                resource = builder.AddResource(project)
+                    .WithAnnotation(instance!)
+                    .WithAnnotation(new TLambdaProject());
+            }
+            else
+            {
```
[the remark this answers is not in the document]
philasmar: Done

src/Aspire.Hosting.AWS/Utils/ProjectUtilities.cs
```diff
@@ -0,0 +1,184 @@
...
+    public static string CreateExecutableWrapperProject(string classLibraryProjectPath, string lambdaHandler)
+    {
+        string tempPath = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString());
+        Directory.CreateDirectory(tempPath);
+
+        var projectContent = $@"
+<Project Sdk=""Microsoft.NET.Sdk"">
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>{Constants.ExecutableWrapperProjectTargetFramework}</TargetFramework>
```
normj, line 159: This shouldn't be a constant. We should use the target framework the Lambda function is using.

src/Aspire.Hosting.AWS/Constants.cs
```diff
@@ -27,4 +27,29 @@ internal static class Constants
...
+    
+    /// <summary>
+    /// The launch settings file name
+    /// </summary>
+    internal const string LaunchSettingsFile = "launchSettings.json";
+    
+    /// <summary>
+    /// The target framework of the executable wrapper project
+    /// </summary>
+    internal const string ExecutableWrapperProjectTargetFramework = "net8.0";
```
normj, line 44: See my comment below that this should not be a constant but we should use the target framework the Lambda project is using.

src/Aspire.Hosting.AWS/Utils/Internal/ProcessCommandService.cs
normj: I'm fine for now since these are internal but eventually we should figure out how to consolidate this.

normj (changes requested):

src/Aspire.Hosting.AWS/Utils/ProjectUtilities.cs
philasmar: Done

PR #48, opened 2025-03-11
Add support for configuring a port for API Gateway emulator

fennekin23:
*Issue #, if available:*
N/A

*Description of changes:*
This change allows overriding the API Gateway and Lambda emulators port. It could be useful in situations where external systems rely on a specific port, for example, during tests.

### DIFF
--- a/src/Aspire.Hosting.AWS/Lambda/APIGatewayExtensions.cs
+++ b/src/Aspire.Hosting.AWS/Lambda/APIGatewayExtensions.cs
@@ -27,6 +27,8 @@ public static class APIGatewayExtensions
     /// <returns></returns>
     public static IResourceBuilder<APIGatewayEmulatorResource> AddAWSAPIGatewayEmulator(this IDistributedApplicationBuilder builder, string name, APIGatewayType apiGatewayType, APIGatewayEmulatorOptions? options = null)
     {
+        options ??= new APIGatewayEmulatorOptions();
+
         var apiGatewayEmulator = builder.AddResource(new APIGatewayEmulatorResource(name, apiGatewayType)).ExcludeFromManifest();
         apiGatewayEmulator.WithArgs(context =>
         {
@@ -36,7 +38,7 @@ public static class APIGatewayExtensions
         var annotation = new EndpointAnnotation(
             protocol: ProtocolType.Tcp,
             uriScheme: "http",
-            port: options?.Port);
+            port: options.Port);
 
         apiGatewayEmulator.WithAnnotation(annotation);
         var endpointReference = new EndpointReference(apiGatewayEmulator.Resource, annotation);
--- a/src/Aspire.Hosting.AWS/Lambda/APIGatewayExtensions.cs
+++ b/src/Aspire.Hosting.AWS/Lambda/APIGatewayExtensions.cs
@@ -17,14 +17,15 @@ namespace Aspire.Hosting;
 public static class APIGatewayExtensions
 {
     /// <summary>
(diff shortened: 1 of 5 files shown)


src/Aspire.Hosting.AWS/Lambda/APIGatewayExtensions.cs
```diff
@@ -17,14 +17,15 @@ namespace Aspire.Hosting;
...
+    /// Adds an API Gateway emulator resource to the Aspire application. Lambda function resources
     /// should be added to this resource using the WithReference method.
     /// </summary>
     /// <param name="builder"></param>
     /// <param name="name">Aspire resource name</param>
     /// <param name="apiGatewayType">The type of API Gateway API. For example Rest, HttpV1 or HttpV2</param>
+    /// <param name="port">Desired port for the API Gateway emulator. For example 5050</param>
     /// <returns></returns>
-    public static IResourceBuilder<APIGatewayEmulatorResource> AddAWSAPIGatewayEmulator(this IDistributedApplicationBuilder builder, string name, APIGatewayType apiGatewayType)
+    public static IResourceBuilder<APIGatewayEmulatorResource> AddAWSAPIGatewayEmulator(this IDistributedApplicationBuilder builder, string name, APIGatewayType apiGatewayType, int? port)
```
normj, line 28: So we don't have to keep changing the signature whenever we add a new optional feature can you create a `APIGatewayEmulatorOptions` and add the port property that. Then change this method to 
```
public static IResourceBuilder<APIGatewayEmulatorResource> AddAWSAPIGatewayEmulator(this IDistributedApplicationBuilder builder, string name, APIGatewayType apiGatewayType, APIGatewayEmulatorOptions options = null)
```

fennekin23: Nice idea! I made the change you suggested.

src/Aspire.Hosting.AWS/Lambda/APIGatewayExtensions.cs
```diff
@@ -17,15 +17,18 @@ namespace Aspire.Hosting;
 public static class APIGatewayExtensions
 {
     /// <summary>
-    /// Adds an API Gateway emulator resource to the Aspire application. Lambda function resoures 
+    /// Adds an API Gateway emulator resource to the Aspire application. Lambda function resources
```
normj, line 20: Thanks for cleaning up my typos!

normj (approved):
Changes look good. 

Any chance you would be willing to make the same change to the [AddAWSLambdaServiceEmulator](https://github.com/aws/integrations-on-dotnet-aspire-for-aws/blob/main/src/Aspire.Hosting.AWS/Lambda/LambdaExtensions.cs#L123) method?

fennekin23:
> Changes look good.
> 
> Any chance you would be willing to make the same change to the [AddAWSLambdaServiceEmulator](https://github.com/aws/integrations-on-dotnet-aspire-for-aws/blob/main/src/Aspire.Hosting.AWS/Lambda/LambdaExtensions.cs#L123) method?

Extended `LambdaFunctionOptions` with `Port` and explicitly set default values for properties. Made a change to `AddAWSLambdaServiceEmulator` to set a port.

normj (approved):
Looks great, thanks for the PR

GarrettBeatty (approved):

normj:
This PR went out today as part of version `9.1.6` of Aspire.Hosting.AWS. Thanks again for the PR.

Issue #55, opened 2025-04-10
There is already a Construct with name 'QueueUrl' when using both WithReference and WithSQSEventSource


dil-adatoth:
When both `WithReference` and `WithSQSEventSource` is being used in the same Aspire project, an exception is thrown: `There is already a Construct with name 'QueueUrl' in Stack`

A minimal example that results in error:

```c#
var sqs1 = stack.AddSQSQueue("test1");
var sqs2 = stack.AddSQSQueue("test2");
builder.AddProject<Projects.Project1>("Project1")
    .WithReference(sqs1);
builder.AddAWSLambdaFunction<Projects.Project2>("Project2", "Project2::Project2.Function::FunctionHandler")
    .WithSQSEventSource(sqs2);
```

As a workaround, right now we can use the queueUrl parameter of `WithSQSEventSource`.
```c#
builder.AddAWSLambdaFunction<Projects.Project2>("Project2", "Project2::Project2.Function::FunctionHandler")
    .WithSQSEventSource(sqs2.Resource.Construct.QueueUrl));
```

normj:
Thanks for the bug report. I was able to reproduce the problem and investigating a fix.

normj:
PR is out to address the issue: https://github.com/aws/integrations-on-dotnet-aspire-for-aws/pull/56

normj:
Version 9.1.7 of `Aspire.Hosting.AWS` is out with this bug fix. Thanks again for reporting the issue.

dil-adatoth:
Thank you @normj for the quick fix! I can verify that in the case we use one SQS, the registration works now:

```c#
var sqs1 = stack.AddSQSQueue("test1");
builder.AddProject<Projects.Project1>("Project1")
    .WithReference(sqs1);
builder.AddAWSLambdaFunction<Projects.Project2>("Project2", "Project2::Project2.Function::FunctionHandler")
    .WithSQSEventSource(sqs1); // This works now
```

However, the code sample that I attached in the first message - when we use two separate SQS queues - I still get the same error. 

```c#
var sqs1 = stack.AddSQSQueue("test1");
var sqs2 = stack.AddSQSQueue("test2");
builder.AddProject<Projects.Project1>("Project1")
    .WithReference(sqs1);
builder.AddAWSLambdaFunction<Projects.Project2>("Project2", "Project2::Project2.Function::FunctionHandler")
    .WithSQSEventSource(sqs2); // This still throws the exception
```

[...]

PR #59, opened 2025-04-11
Fixed having duplicate CDK output parameter with same name

normj:
*Issue #, if available:*
https://github.com/aws/integrations-on-dotnet-aspire-for-aws/issues/55#issuecomment-2796146879

*Description of changes:*
When we add a CDK output construct to the CDK stack the id of the CDK output construct was generated by the logical name like `QueueUrl` without taking into context the parent construct. So if multiple queues were creating output parameters multiple output constructs with the id of `QueueUrl` was attempted and failed.

The fix is to generate the id of the output construct using the context of the parent construct the same we do for the actual output key name.

### DIFF
--- a/playground/Lambda/Lambda.AppHost/Program.cs
+++ b/playground/Lambda/Lambda.AppHost/Program.cs
@@ -10,7 +10,8 @@ var builder = DistributedApplication.CreateBuilder(args);
 var awsSdkConfig = builder.AddAWSSDKConfig().WithRegion(Amazon.RegionEndpoint.USWest2);
 
 var cdkStackResource = builder.AddAWSCDKStack("AWSLambdaPlaygroundResources");
-var sqsDemoQueue = cdkStackResource.AddSQSQueue("DemoQueue");
+var sqsDemoQueue1 = cdkStackResource.AddSQSQueue("DemoQueue1");
+var sqsDemoQueue2 = cdkStackResource.AddSQSQueue("DemoQueue2");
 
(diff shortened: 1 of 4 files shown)


philasmar (approved):

playground/Lambda/Lambda.AppHost/Program.cs
```diff
@@ -35,10 +36,10 @@
 
 builder.AddAWSLambdaFunction<Projects.SQSProcessorFunction>("SQSProcessorFunction", "SQSProcessorFunction::SQSProcessorFunction.Function::FunctionHandler")
         .WithReference(awsSdkConfig)
-        .WithSQSEventSource(sqsDemoQueue)
+        .WithSQSEventSource(sqsDemoQueue1)
         // This reference is not necessary. It is added to confirm duplicate
         // CDK output parameters are not attempted to be added.
-        .WithReference(sqsDemoQueue);
+        .WithReference(sqsDemoQueue2);
```
GarrettBeatty, line 42: why did we remove sqsDemoQueue from here, wasnt that testing the other scenario before? shouldnt it be both 

`.WithReference(sqsDemoQueue`) // duplicate
`.WithReference(sqsDemoQueue2);` // not duplicate

normj: Good call. I added a reference for both demo1 and demo2 queues to confirm the behavior. In the future I would like to add another project to the playground that would act as the SQS publisher instead of adding these kind of random references.

GarrettBeatty (approved):

Issue #49, opened 2025-03-18
Option to disable validation for AWS credentials


klowdo:
Nice additions where made in #30 
 
But when we run our integrations tests in ci pipeline we don't have a profile and don't want it to reach out to aws at all. 
I get this error 
` [Error](03/18/2025 11:35) 816: 2025-03-18T11:35:18.2065247Z Failed to connect to AWS using AWS SDK config: Profile:,Region:eu-west-1` 
That is showing that we don't have a profile.

I made some workaround like this  

```c#
    .WithReference(awsConfig, noValidation: isIntegrationTests)
```

So something like this would be nice or if you have a smarter solution for this, 

Thanks in advance!

normj:
Sounds like a good idea. Would it be better to turn off the validation on `AddAWSSDKConfig` call? I'm assuming if you want validation turned off you would want it turned off in all references. There could be a `WithValidation` method you could fluently call after the `AddAWSSDKConfig` that will set some state on the sdk config and where the validations are done check that state.

If you are interested in contributing a PR to get the work added sooner I'm happy to take a look at a PR.

normj:
PR was released as part of version `9.1.9` of Aspire.Hosting.AWS.

PR #63, opened 2025-04-16
Add option to disable sdk validation on profile

klowdo:
Issue #49

Description of changes:
Adds option to disable sdk validation in

### DIFF
--- a/src/Aspire.Hosting.AWS/AWSSDKConfig.cs
+++ b/src/Aspire.Hosting.AWS/AWSSDKConfig.cs
@@ -11,4 +11,7 @@ internal sealed class AWSSDKConfig : IAWSSDKConfig
 
     /// <inheritdoc/>
     public RegionEndpoint? Region { get; set; }
+
+    /// <inheritdoc/>
+    public bool SDKValidationEnabled { get; set; } = true;
 }
--- a/src/Aspire.Hosting.AWS/SDKResourceExtensions.cs
+++ b/src/Aspire.Hosting.AWS/SDKResourceExtensions.cs
@@ -47,6 +47,18 @@ public static class SDKResourceExtensions
     }
 
     /// <summary>
+    /// Set validation preference for profile in sdk.
+    /// </summary>
+    /// <param name="config">An <see cref="IAWSSDKConfig"/> instance.</param>
+    /// <param name="sdkValidationEnabled">option for toggle sdk validation</param>
+    public static IAWSSDKConfig WithSdkValidation(this IAWSSDKConfig config, bool sdkValidationEnabled)
+    {
+        config.SDKValidationEnabled = sdkValidationEnabled;
+
+        return config;
+    }
+
+    /// <summary>
     /// Add a reference to an AWS SDK configuration to the resource.
     /// </summary>
     /// <param name="builder">An <see cref="IResourceBuilder{T}"/> for <see cref="IResourceWithEnvironment"/></param>
--- a/src/Aspire.Hosting.AWS/SdkUtilities.cs
+++ b/src/Aspire.Hosting.AWS/SdkUtilities.cs
@@ -45,7 +45,7 @@ internal static class SdkUtilities
 
     internal static void ApplySDKConfig(EnvironmentCallbackContext context, IAWSSDKConfig awsSdkConfig, bool force)
     {
-        if (context.Logger != null)
+        if (context.Logger != null && awsSdkConfig.SDKValidationEnabled)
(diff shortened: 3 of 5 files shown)


GarrettBeatty:
Hi, thanks for the PR! Can you please update the target branch to `dev`. And can you also add a change file, following these steps please https://github.com/aws/integrations-on-dotnet-aspire-for-aws/blob/dev/CONTRIBUTING.md#adding-a-change-file-to-your-contribution-branch? Thanks!

src/Aspire.Hosting.AWS/AWSSDKConfig.cs
```diff
@@ -11,4 +11,10 @@ internal sealed class AWSSDKConfig : IAWSSDKConfig
 
     /// <inheritdoc/>
     public RegionEndpoint? Region { get; set; }
+
+    /// <summary>
+    /// Sets if the sdk will validate the profile.
+    /// </summary>
+    internal bool DisableSdkValidation { get; set; }
```
GarrettBeatty, line 18: i wonder if we should add this to the IAWSSDKConfig interface? And then update this to be public and also have ///inheritdoc/> similar to RegionEndpoint

src/Aspire.Hosting.AWS/SDKResourceExtensions.cs
```diff
@@ -46,6 +46,24 @@ public static IAWSSDKConfig WithRegion(this IAWSSDKConfig config, RegionEndpoint
         return config;
     }
 
+    /// <summary>
+    /// Disable validation for profile in sdk.
+    /// </summary>
+    /// <param name="config">An <see cref="IAWSSDKConfig"/> instance.</param>
+    public static IAWSSDKConfig WithNoValidation(this IAWSSDKConfig config)
+    {
+        if (config is AWSSDKConfig awsConfig)
```
GarrettBeatty, line 55: check my above comment - i think if we move that property to the IAWSSDKConfig interface then we wont need this cast. and the function can simply be

```
config.SDKValidationEnabled= enabled;
return config;
```

src/Aspire.Hosting.AWS/SDKResourceExtensions.cs
```diff
@@ -46,6 +46,24 @@ public static IAWSSDKConfig WithRegion(this IAWSSDKConfig config, RegionEndpoint
         return config;
     }
 
+    /// <summary>
+    /// Disable validation for profile in sdk.
+    /// </summary>
+    /// <param name="config">An <see cref="IAWSSDKConfig"/> instance.</param>
+    public static IAWSSDKConfig WithNoValidation(this IAWSSDKConfig config)
```
GarrettBeatty, line 53: I think we should make this `WithValidation` and then take in a `enabled` parameter. Rather than having WithNoValidation, make it the inverse - i think it will be easier to read.

Then the user can call `WithValidation(false)`

src/Aspire.Hosting.AWS/SdkUtilities.cs
```diff
@@ -45,7 +45,8 @@ internal static void ConfigureUserAgentString(object sender, RequestEventArgs e)
 
     internal static void ApplySDKConfig(EnvironmentCallbackContext context, IAWSSDKConfig awsSdkConfig, bool force)
     {
-        if (context.Logger != null)
+        var disableSdkValidation = (awsSdkConfig as AWSSDKConfig)?.DisableSdkValidation ?? false;
+        if (context.Logger != null && !disableSdkValidation)
```
GarrettBeatty, line 49: if we update the logic to be inverted (SDKValidationEnabled), then update this area too

src/Aspire.Hosting.AWS/AWSSDKConfig.cs
GarrettBeatty: Also i think we should make this the inverse to make it easier to read - SDKValidationEnabled  and have the default value as `true`

GarrettBeatty (changes requested):

src/Aspire.Hosting.AWS/SDKResourceExtensions.cs
klowdo: I think `WithSDKValidation` is better. who knows what validation will come to this in the future :)

GarrettBeatty (approved):
looks good to me! Thank you for this!

klowdo:
Thank you!

philasmar (approved):

PR #89, opened 2025-06-06
Add the option for users to specify custom cloudformation tags

GarrettBeatty:
*Issue #, if available:*
DOTNET-8128

*Description of changes:*
1. Add the option for users to specify custom cloudformation tags

Example usage
```
var awsResources = builder.AddAWSCloudFormationTemplate("AspireSampleDevResources", "app-resources.template")
                          .WithParameter("DefaultVisibilityTimeout", "30")
                          .WithReference(awsConfig)
                          .WithTag("Environment", "Test")
                          .WithTag("Project", "AspireSample")
                          .WithTag("Owner", "DotNetTeam");
```

```
var customStack = builder.AddAWSCDKStack("custom", scope => new CustomStack(scope, "Aspire-custom"));
customStack.AddOutput("BucketName", stack => stack.Bucket.BucketName).WithReference(awsConfig);
customStack.WithTag("Environment", "Test");
```

In below picture i deployed a stack with `WithTag("Environment", "Test")`

image

### DIFF
--- a/src/Aspire.Hosting.AWS/Provisioning/CloudFormationStackExecutor.cs
+++ b/src/Aspire.Hosting.AWS/Provisioning/CloudFormationStackExecutor.cs
@@ -62,20 +62,34 @@ internal sealed class CloudFormationStackExecutor(
     }
 
     /// <summary>
-    /// Setup the tags collection by coping the any tags for existing stacks and then updating/adding the tag recording
-    /// the SHA256 for template and parameters.
+    /// Setup the tags collection by copying any tags for existing stacks, adding user-specified tags, 
+    /// and then updating/adding the tag recording the SHA256 for template and parameters.
     /// </summary>
     /// <param name="existingStack"></param>
     /// <param name="changeSetType"></param>
     /// <param name="computedSha256"></param>
     /// <returns></returns>
-    private static (List<Tag> tags, string? existingSha256) SetupTags(Stack? existingStack, ChangeSetType changeSetType, string computedSha256)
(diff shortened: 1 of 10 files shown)


src/Aspire.Hosting.AWS/Provisioning/CloudFormationStackExecutor.cs
```diff
@@ -78,6 +78,20 @@ private static (List<Tag> tags, string? existingSha256) SetupTags(Stack? existin
             tags = existingStack.Tags ?? new List<Tag>();
         }
 
+        // Add user-specified tags from context
+        foreach (var kvp in context.Tags)
```
normj, line 82: In an update scenario we are starting with a collection of tags filled with the existing tags would this logic work if in the AppHost I deployed with a tag and then want to remove the tag?

normj (commented):
Seems like the CDK version of provisioning needs tagging support. I think in CDK you have the tags collection on the CDK stack object but I'm pretty sure our provisioning logic will ignore it.

I'm okay if you want to separate the CDK work in a separate effort but if you do can you add a story in the backlog so we don't lose it.

src/Aspire.Hosting.AWS/Provisioning/CloudFormationStackExecutor.cs
GarrettBeatty: ill have to test this scenario

GarrettBeatty: actually ignore this comment. my latest commit i was still having other hardcoded stuff in it. so i need to double check again

Issue #104, opened 2025-06-25
NullReferenceException when provisioning Cloud Foundation template without any outputs


EtherZa:
`CloudFormationTemplateResourceProvisioner` throws a `NullReferenceException` when provisioning a template that does not include any outputs.

![Image](https://github.com/user-attachments/assets/27d4f53d-e70e-4855-81d0-d433aab36f5d)

normj:
Version `9.2.5` of Aspire.Hosting.AWS is out with your PR. Thanks for the fix!

EtherZa:
That was quick. Thank you!

PR #105, opened 2025-06-25
Fixed NullReferenceException thrown when provisioning a Cloud Foundat…

EtherZa:
CloudFormationTemplateResourceProvisioner throws a NullReferenceException when provisioning a template that does not include any outputs. The exception occurs when attempting to log the number out outputs.

Issue: 
* #104 

Description of changes:
* Modified the log argument to use a null-coalesce with a fallback of 0 to prevent the exception.


normj (approved):
Thanks for the PR addressing the issue.

PR #107, opened 2025-06-26
Integrate Lambda Test Tool's feature for allowing saving requests

normj:
*Description of changes:*
Upgrade to version `0.11.0` of Amazon.Lambda.TestTool` and integrate it's new feature of allowing saving requests from the test tool. By default the test tool will enable saving requests in the `<current directory>/.aws-lambda-testool`. The current directory most likely will be the AppHost. The location can be overridden by explicitly calling the `AddAWSLambdaServiceEmulator` with an `LambdaEmulatorOptions` instance and setting the `ConfigStoragePath` property. If users want to disable saving feature they can set the `ConfigStoragePath` property to empty string. I can't rely on a `null` check for `ConfigStoragePath` because by default the containing `LambdaEmulatorOptions` object is null. So it would get tricky to tell the difference of `ConfigStoragePath` being null because user explicitly set it or the containing object never existed. 

Tested by manually confirming the expected behavior in the playground app and adding new integ tests.

### DIFF
--- a/src/Aspire.Hosting.AWS/Constants.cs
+++ b/src/Aspire.Hosting.AWS/Constants.cs
@@ -41,10 +41,15 @@ internal static class Constants
     /// <summary>
     /// The version of RuntimeSupport used in the executable wrapper project
     /// </summary>
-    internal const string RuntimeSupportPackageVersion = "1.13.0";
+    internal const string RuntimeSupportPackageVersion = "1.13.1";
     
     /// <summary>
     /// The default version of Amazon.Lambda.TestTool that will be automatically installed
     /// </summary>
-    internal const string DefaultLambdaTestToolVersion = "0.10.3";
+    internal const string DefaultLambdaTestToolVersion = "0.11.0";
+
+    /// <summary>
+    /// The default directory the Lambda Test Tool will be configured for storing configuration information like saved requests.
+    /// </summary>
+    internal const string DefaultLambdaConfigStorage = ".aws-lambda-testool";
 }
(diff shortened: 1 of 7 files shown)


src/Aspire.Hosting.AWS/Constants.cs
```diff
@@ -41,10 +41,15 @@ internal static class Constants
...
     /// <summary>
     /// The default version of Amazon.Lambda.TestTool that will be automatically installed
     /// </summary>
-    internal const string DefaultLambdaTestToolVersion = "0.10.3";
+    internal const string DefaultLambdaTestToolVersion = "0.11.0";
+
+    /// <summary>
+    /// The default directory the Lambda Test Tool will be configured for storing configuration information like saved requests.
+    /// </summary>
+    internal const string DefaultLambdaConfigStorage = ".aws-lambda-testool";
```
philasmar, line 54: Typo in testtool

philasmar (approved):

src/Aspire.Hosting.AWS/Constants.cs
normj: Fixed

tests/Aspire.Hosting.AWS.Integ.Tests/Lambda/PlaygroundE2ETests.cs
```diff
@@ -1,10 +1,12 @@
...
-using Amazon;
+﻿using Amazon;
 using Amazon.CloudFormation;
 using Amazon.CloudFormation.Model;
 using Amazon.SQS;
 using Amazon.SQS.Model;
+using Aspire.Hosting.AWS.CDK;
+using Aspire.Hosting.AWS.Lambda;
+
+#pragma warning disable CA2252
```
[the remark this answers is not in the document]
normj: Because the new tests I added are calling directly our `AddAWSLambdaServiceEmulator` method that we have marked as preview. This is the same as you can find in our playground app.

Issue #136, opened 2025-11-08
API Gateway Emulator exposes only HTTP endpoint while Lambda handler reports only HTTPS


nazarii-piontko:
### Describe the bug

When using **.NET Minimal API** hosted as **AWS Lambda** and running locally with **Aspire**, there is a mismatch between the protocol used by the API Gateway emulator and what the Lambda handler reports.

The emulator created with `AddAWSAPIGatewayEmulator` exposes an **HTTP** endpoint. However, the Lambda function configured with `AddAWSLambdaHosting` always reports its request scheme as **HTTPS**.
As a result, any URLs generated within the API (for example, using `Request.Scheme`) incorrectly use the HTTPS schema, even though the emulator only supports HTTP during local development.

This leads to broken URLs and inconsistencies during local development or when using services that depend on correct callback URLs.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

* The `AddAWSAPIGatewayEmulator` should expose an **HTTPS** endpoint (or at least support an HTTPS mode) so that the scheme matches what the Lambda hosting reports.
* Alternatively, the Lambda hosting component (`AddAWSLambdaHosting`) should correctly detect and align with the emulator’s actual protocol.

### Current Behavior

* The emulator exposes an **HTTP** endpoint.
* The Lambda handler reports **HTTPS** for all requests.
* Generated URLs contain incorrect schemas (`https://` instead of `http://`), leading to broken links and local development issues.

### Reproduction Steps

1. Create a .NET Minimal API project configured for AWS Lambda using `AddAWSLambdaHosting`.
2. Configure the Aspire host, for example:

   ```csharp
   builder.AddAWSLambdaFunction<Projects.Api>("api", "Api");
   builder.AddAWSAPIGatewayEmulator("api-gateway-emulator", APIGatewayType.HttpV2).WithReference(api, Method.Any, "/{proxy+}")
   ```
3. Run the Aspire solution locally.
4. Call any endpoint that returns or constructs a URL (in my case, it is OpenAPI - `MapOpenApi`).
5. Observe that:

   * The emulator serves on `http://localhost:xxxx`.
   * The reported scheme inside the Lambda handler is `https`.

[AspireWithAwsIssues.zip](https://github.com/user-attachments/files/23434655/AspireWithAwsIssues.zip)

### Possible Solution

[...]

PR #156, opened 2026-02-11
Integrate Lambda Test Tool's HTTPS support

normj:
## Description
To more closely match Lambda and more importantly API Gateway experience expose HTTPS endpoints for the emulator. This PR is contingent on [PR](https://github.com/aws/aws-lambda-dotnet/pull/2282) for the test tool being released that adds support for enabling HTTPS support.

Aspire itself takes care of provisioning dev certs for the Aspire resource and sets the `SSL_CERT_DIR` environment variable for the ASP.NET Core part of the Lambda Test Tool to pick up and automatically use.

The user experience will be that HTTPS endpoints will provisioned by default for users. If users don't want to HTTPS endpoints they can set the `DisableHttpsEndpoint` property. A reason I can think a user wanting to do this is if their environment doesn't allow the certs that Aspire provisions. The HTTPS port can be configured with the new `HttpsPort` port, if not set a random port will be allocated via Aspire. **Breaking Change** the existing `Port` property was renamed to `HttpPort` to match the new `HttpsPort` property.

## Motivation and Context
https://github.com/aws/integrations-on-dotnet-aspire-for-aws/issues/136

## Testing
Testing with HTTPS ports through xunit is tricky and I doubt our CI system will have certs in place so I relied on manually testing building the test tool into NuGet packages and testing it directly along with the Aspire integration.

### DIFF
--- a/src/Aspire.Hosting.AWS/Lambda/APIGatewayEmulatorOptions.cs
+++ b/src/Aspire.Hosting.AWS/Lambda/APIGatewayEmulatorOptions.cs
@@ -8,7 +8,17 @@ namespace Aspire.Hosting.AWS.Lambda;
 public class APIGatewayEmulatorOptions
 {
     /// <summary>
-    /// The port that the API Gateway emulator will listen on. If not set, a random port will be used.
+    /// The http port that the API Gateway emulator will listen on. If not set, a random port will be used.
     /// </summary>
-    public int? Port { get; set; } = null;
+    public int? HttpPort { get; set; } = null;
+
+    /// <summary>
+    /// The https port that the API Gateway emulator will listen on. If not set, a random port will be used.
+    /// </summary>
+    public int? HttpsPort { get; set; } = null;
+
+    /// <summary>
+    /// Setting to true will disable the creation of the HTTPS endpoint for the API Gateway emulator.
+    /// </summary>
+    public bool DisableHttpsEndpoint { get; set; } = false;
 }
--- a/src/Aspire.Hosting.AWS/Lambda/APIGatewayEmulatorOptions.cs
+++ b/src/Aspire.Hosting.AWS/Lambda/APIGatewayEmulatorOptions.cs
@@ -8,12 +8,12 @@ namespace Aspire.Hosting.AWS.Lambda;
 public class APIGatewayEmulatorOptions
 {
     /// <summary>
-    /// The http port that the API Gateway emulator will listen on. If not set, a random port will be used.
+    /// The HTTP port that the API Gateway emulator will listen on. If not set, a random port will be used.
     /// </summary>
     public int? HttpPort { get; set; } = null;
 
     /// <summary>
-    /// The https port that the API Gateway emulator will listen on. If not set, a random port will be used.
+    /// The HTTPS port that the API Gateway emulator will listen on. If not set, a random port will be used.
     /// </summary>
     public int? HttpsPort { get; set; } = null;
 
--- a/src/Aspire.Hosting.AWS/Lambda/APIGatewayExtensions.cs
+++ b/src/Aspire.Hosting.AWS/Lambda/APIGatewayExtensions.cs
@@ -35,20 +35,41 @@ public static class APIGatewayExtensions
             apiGatewayEmulator.Resource.AddCommandLineArguments(context.Args);
         });
 
-        var annotation = new EndpointAnnotation(
+        var annotationHttp = new EndpointAnnotation(
             protocol: ProtocolType.Tcp,
             uriScheme: "http",
-            port: options.Port);
+            port: options.HttpPort);
+        apiGatewayEmulator.WithAnnotation(annotationHttp);
+        var endpointHttpReference = new EndpointReference(apiGatewayEmulator.Resource, annotationHttp);
 
-        apiGatewayEmulator.WithAnnotation(annotation);
-        apiGatewayEmulator.WithUrlForEndpoint("http", u => u.DisplayText = "API Gateway Endpoint");
+        EndpointReference? endpointHttpsReference = null;
 
-        var endpointReference = new EndpointReference(apiGatewayEmulator.Resource, annotation);
+        if (!options.DisableHttpsEndpoint)
+        {
+            var annotationHttps = new EndpointAnnotation(
+                protocol: ProtocolType.Tcp,
+                uriScheme: "https",
+                port: options.HttpsPort);
+            apiGatewayEmulator.WithAnnotation(annotationHttps);
+            endpointHttpsReference = new EndpointReference(apiGatewayEmulator.Resource, annotationHttps);
+
+            apiGatewayEmulator.WithUrlForEndpoint("https", u => u.DisplayText = "API Gateway Endpoint");
+        }
+        else
+        {
+            apiGatewayEmulator.WithUrlForEndpoint("http", u => u.DisplayText = "API Gateway Endpoint");
+        }
+
 
         apiGatewayEmulator.WithAnnotation(new EnvironmentCallbackAnnotation(context =>
         {
             context.EnvironmentVariables[Constants.IsAspireHostedEnvVariable] = "true";
-            context.EnvironmentVariables["API_GATEWAY_EMULATOR_PORT"] = endpointReference.Property(EndpointProperty.TargetPort);
+            context.EnvironmentVariables["API_GATEWAY_EMULATOR_PORT"] = endpointHttpReference.Property(EndpointProperty.TargetPort);
+
+            if (endpointHttpsReference != null)
+            {
+                context.EnvironmentVariables["API_GATEWAY_EMULATOR_HTTPS_PORT"] = endpointHttpsReference.Property(EndpointProperty.TargetPort);
+            }
         }));
 
         apiGatewayEmulator.WithAnnotation(new APIGatewayEmulatorAnnotation(apiGatewayType));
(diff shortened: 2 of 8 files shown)


src/Aspire.Hosting.AWS/Lambda/LambdaEmulatorAnnotation.cs
```diff
@@ -7,13 +7,13 @@ namespace Aspire.Hosting.AWS.Lambda;
...
 /// Annotation for the metadata of a Lambda runtime emulator resource
 /// </summary>
-/// <param name="endpoint"></param>
-internal class LambdaEmulatorAnnotation(EndpointReference endpoint) : IResourceAnnotation
+/// <param name="lambdaRuntimeEndpoint"></param>
+internal class LambdaEmulatorAnnotation(EndpointReference lambdaRuntimeEndpoint) : IResourceAnnotation
 {
-    /// <summary>
-    /// The HTTP endpoint for the Lambda runtime emulator.
+    /// <summary>new
```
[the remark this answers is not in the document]
normj: Fixed

src/Aspire.Hosting.AWS/Lambda/APIGatewayExtensions.cs
```diff
@@ -35,20 +35,41 @@ public static IResourceBuilder<APIGatewayEmulatorResource> AddAWSAPIGatewayEmula
...
         apiGatewayEmulator.WithAnnotation(new EnvironmentCallbackAnnotation(context =>
         {
             context.EnvironmentVariables[Constants.IsAspireHostedEnvVariable] = "true";
-            context.EnvironmentVariables["API_GATEWAY_EMULATOR_PORT"] = endpointReference.Property(EndpointProperty.TargetPort);
+            context.EnvironmentVariables["API_GATEWAY_EMULATOR_PORT"] = endpointHttpReference.Property(EndpointProperty.TargetPort);
+
+            if (endpointHttpsReference != null)
+            {
+                context.EnvironmentVariables["API_GATEWAY_EMULATOR_HTTPS_PORT"] = endpointHttpsReference.Property(EndpointProperty.TargetPort);
+            }
```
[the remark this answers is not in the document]
normj: Added tests

src/Aspire.Hosting.AWS/Lambda/LambdaExtensions.cs
```diff
@@ -157,7 +179,12 @@ public static IResourceBuilder<LambdaEmulatorResource> AddAWSLambdaServiceEmulat
...
         {
             context.EnvironmentVariables[Constants.IsAspireHostedEnvVariable] = "true";
-            context.EnvironmentVariables["LAMBDA_RUNTIME_API_PORT"] = endpointReference.Property(EndpointProperty.TargetPort);
+            context.EnvironmentVariables["LAMBDA_RUNTIME_API_PORT"] = endpointHttpReference.Property(EndpointProperty.TargetPort);
+
+            if (!options.DisableHttpsEndpoint && endpointHttpsReference != null)
+            {
+                context.EnvironmentVariables["LAMBDA_WEB_UI_HTTPS_PORT"] = endpointHttpsReference.Property(EndpointProperty.TargetPort);
+            }
         }));
```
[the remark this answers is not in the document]
normj: Added tests

src/Aspire.Hosting.AWS/Lambda/APIGatewayExtensions.cs
```diff
@@ -35,20 +35,41 @@ public static IResourceBuilder<APIGatewayEmulatorResource> AddAWSAPIGatewayEmula
...
+            apiGatewayEmulator.WithUrlForEndpoint("http", u => u.DisplayText = "API Gateway Endpoint");
+        }
 
-        var endpointReference = new EndpointReference(apiGatewayEmulator.Resource, annotation);
 
         apiGatewayEmulator.WithAnnotation(new EnvironmentCallbackAnnotation(context =>
         {
             context.EnvironmentVariables[Constants.IsAspireHostedEnvVariable] = "true";
-            context.EnvironmentVariables["API_GATEWAY_EMULATOR_PORT"] = endpointReference.Property(EndpointProperty.TargetPort);
+            context.EnvironmentVariables["API_GATEWAY_EMULATOR_PORT"] = endpointHttpReference.Property(EndpointProperty.TargetPort);
```
GarrettBeatty, line 67: should we also rename this to API_GATEWAY_EMULATOR_HTTP_PORT similar to how we made the breaking change for the other variable?

GarrettBeatty (approved):

src/Aspire.Hosting.AWS/Lambda/APIGatewayExtensions.cs
normj: I was on the fence on the test tool PR whether to make breaking changes to the environment variables and command line switches. First I think the environment variable name should match the switch name so if we changed the environment variable we should making the breaking change to the switch.

From the view point of the test tool only I think using HTTPS is harder because some how the certs have to be setup which in the aspire use case is handled for us. But if you are using the test tool standalone you are going to need to make sure dev certs have been setup. So my thinking was for the test tool HTTP should always be viewed as the default hence it was okay just calling it port instead of http port. Using https is the advanced use case and should be viewed as the extra effort not the equal as just setting the port. Plus with Lambda emulator side there is the complexity that port is always required for the Lambda Runtime API.

I'm not sure I'm 100% right so welcome for different view. Also curious what @philasmar take is.

philasmar: I don't think we should make the breaking change to the http port. I'd say we should leave it as is, for the switch name and env var name. By default, `Port` should be HTTP.

src/Aspire.Hosting.AWS/Lambda/APIGatewayEmulatorOptions.cs
```diff
@@ -8,7 +8,17 @@ namespace Aspire.Hosting.AWS.Lambda;
 public class APIGatewayEmulatorOptions
 {
     /// <summary>
-    /// The port that the API Gateway emulator will listen on. If not set, a random port will be used.
+    /// The http port that the API Gateway emulator will listen on. If not set, a random port will be used.
```
philasmar: nit: HTTP should be uppercase

src/Aspire.Hosting.AWS/Lambda/APIGatewayEmulatorOptions.cs
```diff
@@ -8,7 +8,17 @@ namespace Aspire.Hosting.AWS.Lambda;
...
 {
     /// <summary>
-    /// The port that the API Gateway emulator will listen on. If not set, a random port will be used.
+    /// The http port that the API Gateway emulator will listen on. If not set, a random port will be used.
     /// </summary>
-    public int? Port { get; set; } = null;
+    public int? HttpPort { get; set; } = null;
+
+    /// <summary>
+    /// The https port that the API Gateway emulator will listen on. If not set, a random port will be used.
```
philasmar: nit: HTTPS should be uppercase

src/Aspire.Hosting.AWS/Lambda/LambdaEmulatorOptions.cs
```diff
@@ -28,13 +28,24 @@ public class LambdaEmulatorOptions
     public bool AllowDowngrade { get; set; } = false;
 
     /// <summary>
-    /// The port that the Lambda emulator will listen on. If not set, a random port will be used.
+    /// The http port that the Lambda emulator will listen on. If not set, a random port will be used.
```
philasmar: nit: HTTP should be uppercase

src/Aspire.Hosting.AWS/Lambda/LambdaEmulatorOptions.cs
```diff
@@ -28,13 +28,24 @@ public class LambdaEmulatorOptions
...
     /// </summary>
-    public int? Port { get; set; } = null;
+    public int? HttpPort { get; set; } = null;
+
+    /// <summary>
+    /// The https port that the Lambda emulator will listen on. If not set, a random port will be used.
```
philasmar: nit: HTTPS should be uppercase

src/Aspire.Hosting.AWS/Lambda/LambdaEmulatorOptions.cs
philasmar: don't think we should rename the switch

src/Aspire.Hosting.AWS/Lambda/APIGatewayEmulatorOptions.cs
philasmar: don't think we should rename the switch

normj: Fixed

src/Aspire.Hosting.AWS/Lambda/LambdaEmulatorOptions.cs
normj: Fixed

src/Aspire.Hosting.AWS/Lambda/APIGatewayEmulatorOptions.cs
normj: Fixed

src/Aspire.Hosting.AWS/Lambda/LambdaEmulatorOptions.cs
normj: Fixed

philasmar (approved):

PR #158, opened 2026-02-21
Add checks for dev cert when choosing to enable Lambda HTTPS ports by default

normj:
## Description
In PR https://github.com/aws/integrations-on-dotnet-aspire-for-aws/pull/156 we enabled HTTPS ports by default for the Lambda and API Gateway emulation. But if a user doesn't have the dev certs for localhost installed, commonly in CI systems, then the emulators will fail to launch.

This PR checks to see of a dev cert is installed and if not changes the default experience to not have HTTPS endpoints.

## Motivation and Context
https://github.com/aws/integrations-on-dotnet-aspire-for-aws/issues/136#issuecomment-3932936805

## Testing
I had to rely on manual testing since installing and uninstalling certs is not something I want to do as part of xunit tests. I verified correct behavior with and without the dev certs installed. In the case of not having the dev certs the emulators default to HTTP only. If the dev cert was installed both the HTTPS and HTTP endpoints were allocated.

### DIFF
--- a/src/Aspire.Hosting.AWS/Utils/DevCertificateDetector.cs
+++ b/src/Aspire.Hosting.AWS/Utils/DevCertificateDetector.cs
@@ -1,0 +1,70 @@
+﻿// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+using System.Security.Cryptography.X509Certificates;
+
+namespace Aspire.Hosting.AWS.Utils;
+
+internal class DevCertificateDetector
+{
+    private const string LocalhostSubject = "CN=localhost";
+
+    // This is the value for the Server Authentication EKU using TLS Web Server Authentication, which is required for ASP.NET Core's development certificate.
+    private const string ServerAuthOid = "1.3.6.1.5.5.7.3.1";
+
+    public static bool HasAspNetCoreDevCert()
+    {
+        try
+        {
+            using var store = new X509Store(StoreName.My, StoreLocation.CurrentUser);
+            store.Open(OpenFlags.ReadOnly);
+
+            var now = DateTimeOffset.UtcNow;
+
+            foreach (var cert in store.Certificates)
+            {
+                if (!IsLocalhost(cert))
+                    continue;
+
+                if (!IsValid(cert, now))
+                    continue;
+
+                if (!HasServerAuthEku(cert))
+                    continue;
+
+                if (!cert.HasPrivateKey)
+                    continue;
+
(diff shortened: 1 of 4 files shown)


src/Aspire.Hosting.AWS/Utils/DevCertificateDetector.cs
```diff
@@ -0,0 +1,70 @@
...
+            // Never throw from detection logic in a public library
+            return false;
+        }
+    }
+
+    private static bool IsLocalhost(X509Certificate2 cert) =>
+        string.Equals(cert.Subject, LocalhostSubject, StringComparison.OrdinalIgnoreCase);
+
+    private static bool IsValid(X509Certificate2 cert, DateTimeOffset now) =>
+        now >= cert.NotBefore && now <= cert.NotAfter;
```
boblodgett: You got UtcNow and passed it in but NotBefore and NotAfter appear to be in local time. 

https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509certificate2?view=net-10.0

NotAfter	
Gets the date in local time after which a certificate is no longer valid.

NotBefore	
Gets the date in local time on which a certificate becomes valid.

src/Aspire.Hosting.AWS/Utils/DevCertificateDetector.cs
```diff
@@ -0,0 +1,70 @@
...
+                // Prefer the ASP.NET Core OID if present,
+                // but don't require it for forward compatibility.
+                return true;
+            }
+
+            return false;
+        }
+        catch
+        {
+            // Never throw from detection logic in a public library
```
boblodgett, line 45: Should this at least log something? Cannot tell the difference between this and the normal false.

boblodgett (changes requested):

src/Aspire.Hosting.AWS/Utils/DevCertificateDetector.cs
normj: Unfortunately at the point this is being run the logging system hasn't been setup yet for the Aspire AppHost. And I don't think users would like us bluntly writing to console instead of using logging.

src/Aspire.Hosting.AWS/Utils/DevCertificateDetector.cs
normj: Fixed

boblodgett (approved):

GarrettBeatty (approved):

PR #165, opened 2026-03-12
Fix duplicated resource name when adding more than one SQS Event Source to a resource

msmourao:
Description
This PR adds two improvements to the SQS event source for Lambda:

Resource name customization (SQSEventSourceOptions.ResourceName): New optional property to set a custom resource name for the SQSEventSourceResource. When the same Lambda is wired to multiple SQS queues, this avoids duplicate resource name conflicts in the Aspire dashboard.

Resource name length enforcement: Aspire limits resource names to 64 characters. When the default pattern SQSEventSource-{lambdaName}-{queueName} exceeds this limit (common with long queue names), the code now shortens the name by appending the first 8 characters of a SHA256 hash of the queue name instead of truncating, keeping uniqueness across queues.

Files changed:

SQSEventSourceOptions.cs: Added optional ResourceName property.
SQSEventSourceExtensions.cs: Use ResourceName when set, otherwise generate the default pattern and apply EnsureResourceNameLength() when the name would exceed 64 characters.
Motivation and Context
Duplicate resource names: When multiple SQS queues are attached to one Lambda via WithSQSEventSource, all event sources had the same resource name (e.g. SQSEventSource-{lambdaName}), causing conflicts in the Aspire resource model.
64-character limit: Long queue names (e.g. message-type-based queue names) produce names over Aspire’s 64-character limit and previously could fail or behave incorrectly when truncated.
This change supports modular architectures where one Lambda handles several message types from different queues, which is a standard pattern for event-driven applications.

Testing
Unit tests: Existing SQSEventSourceConfigTests still pass; ResourceName is not part of the config string and is only used for Aspire resource naming.
Integration: Ran the Lambda playground with multiple SQS event sources attached to the same Lambda; event sources are created with distinct resource names and the dashboard shows them correctly.
Length handling: Verified that names longer than 64 characters are shortened via hash; multiple long-named queues produce different hashes and remain unique.
Breaking Changes Assessment
Functionality changes:

SQSEventSourceOptions gains an optional ResourceName property.
Resource name generation logic changed: when a queue name is available, it uses SQSEventSource-{lambdaName}-{queueName} instead of a generic pattern.
Customer impact:

Low: ResourceName is optional; existing code continues to work and benefits from the new default naming.
If customers relied on the old resource names (e.g. SQSEventSource-{lambdaName}) for tests or automation, they may need to update those expectations; they can restore the old behavior by setting ResourceName explicitly.
Reason for change:

Resolves duplicate resource name errors when multiple queues are attached to one Lambda and ensures names stay within Aspire’s 64-character limit.
Best practices: The change follows existing patterns in the codebase and keeps behavior backward compatible via optional configuration.

Testing: Tested locally with the Lambda playground and in a custom AppHost project using multiple SQS event sources on a single Lambda.

Review: A senior/experienced engineer has been assigned to review this PR.


msmourao (commented):
Implemented suggested changes

normj:
Thanks for the PR. This seems like something we want to fix. I'm fixing some CI issues first and then will take a deeper look.

.github/PULL_REQUEST_TEMPLATE.md
normj, line 15: Not sure why git is showing this as a change. It is the same content that is in `main` and `dev`. I'm assuming it will work itself out when the PR is merged.

normj (approved):

philasmar (approved):

Issue #167, opened 2026-03-24
AddAWSLambdaFunction no longer works for Aspire.AppHost.Sdk/13.2.0


Minnbrit:
### Describe the bug

We have an existing AWS Lambda project that we run locally using Aspire.

After upgrading the Aspire.AppHost.Sdk from 13.1.2 -> 13.2.0 the lambda no longer runs.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

When I launch the Aspire AppHost project and I have a breakpoint in the Function constructor.

- The breakpoint gets hit 
- the Aspire Dashboard loads showing that the LambdaServiceEmulator and our lambda are up and running

<img width="615" height="126" alt="Image" src="https://github.com/user-attachments/assets/a3852d9e-2a84-46fd-9d1f-994a14ca26d5" />

- In the lambda test tool I can see our saved requests

<img width="449" height="187" alt="Image" src="https://github.com/user-attachments/assets/8fb514af-7278-4f6a-bca7-e3117fa32242" />

### Current Behavior

- The breakpoint in the function constructor does not get hit
- The Aspire dashboard loads and shows the LambdaServiceEmulator is running by our lambda has stopped

<img width="517" height="128" alt="Image" src="https://github.com/user-attachments/assets/df5b0e86-6fde-4a73-9f63-d716001f3dd1" />

- In the lambda test tool our saved queries do not appear

<img width="488" height="131" alt="Image" src="https://github.com/user-attachments/assets/80fca7e7-d1ab-40db-a1ec-aa6d462b455c" />

- The console logs show these (after the first one they are on stderr)
```
 AWS SDK config validated for account: ...
 Unable to proceed with project '...\ApplicationCleanupLambda.csproj'.
 Ensure you have a runnable project type.
 A runnable project should target a runnable TFM (for instance, net10.0) and have OutputType 'Exe'.
 The current OutputType is 'Library'.
```

### Reproduction Steps

In the AppHost project change
`<Project Sdk="Aspire.AppHost.Sdk/13.1.2">`
to 
`<Project Sdk="Aspire.AppHost.Sdk/13.2.0">`

The AppHost.cs code looks like this

```
using Amazon;

var builder = DistributedApplication.CreateBuilder(args);

var awsAccount= builder.AddAWSSDKConfig()
    .WithRegion(RegionEndpoint.EUWest1)
    .WithProfile("our_profile");

builder
[...]

PR #168, opened 2026-03-27
Fix issue with Lambda functions failing to start when launched in Visual Studio

normj:
## Description
When updating to version 13.2.0 of `Aspire.Hosting` class library based Lambda functions failed to start. Looking through the changes in Aspire 13.2.0 it looks like there were new requirements added to make sure the `SupportsDebuggingAnnotation` attribute is applied for the .NET projects. This is normally not a big deal but in our Lambda's case we don't use the `AddProject` extension method for adding the Lambda projects where the `SupportsDebuggingAnnotation` would normally be added. We can't use that because we need to create a subclass of the `ProjectResource`.

Based on examples I saw in the Aspire repository I got the `SupportsDebuggingAnnotation`  added to the Lambda projects and then debugging worked correctly. 

My understanding without the attribute Aspire falls back to essentially doing a `dotnet run` on the project which fails because the project is a class library.

## Motivation and Context
https://github.com/aws/integrations-on-dotnet-aspire-for-aws/issues/167

## Testing
Launched the dashboard and confirmed the functions were launched and I could hit break points when invoking the functions.

### DIFF
--- a/src/Aspire.Hosting.AWS/Lambda/LambdaExtensions.cs
+++ b/src/Aspire.Hosting.AWS/Lambda/LambdaExtensions.cs
@@ -10,7 +10,7 @@ using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Hosting;
 using System.Diagnostics;
 using System.Net.Sockets;
-using System.Runtime.Versioning;
+using System.Text.Json.Serialization;
 
 #pragma warning disable IDE0130
 namespace Aspire.Hosting;
@@ -19,7 +19,7 @@ namespace Aspire.Hosting;
 /// Extension methods for adding Lambda functions as Aspire resources.
 /// </summary>
 public static class LambdaExtensions
-{   
+{
     /// <summary>
     /// Add a Lambda function as an Aspire resource.
     /// </summary>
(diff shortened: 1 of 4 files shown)


src/Aspire.Hosting.AWS/Lambda/LambdaExtensions.cs
```diff
@@ -41,9 +41,12 @@ public static class LambdaExtensions
             // If we are running Aspire through an IDE where a debugger is attached,
             // we want to configure the Aspire resource to use a Launch Setting Profile that will be able to run the class library Lambda function.
             var project = new LambdaProjectResource(name);
+#pragma warning disable ASPIREEXTENSION001 // WithDebugSupport is experimental
             resource = builder.AddResource(project)
                 .WithAnnotation(new LaunchProfileAnnotation($"{Constants.LaunchSettingsNodePrefix}{name}"))
-                .WithAnnotation(new TLambdaProject());
+                .WithAnnotation(new TLambdaProject())
+                .WithDebugSupport(mode => new LambdaFunctionsLaunchConfiguration { ProjectPath = metadata.ProjectPath, Mode = mode, LaunchProfile = $"{Constants.LaunchSettingsNodePrefix}{name}" }, "project");
```
normj, line 48: I thought for type I should have something like `lambda-function` instead of `project`. That is what I saw for the azure functions https://github.com/microsoft/aspire/blob/45bbcb799bc647bd5ca376b9a8376a65efc2a1dc/src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs#L190. But when I tried that it didn't work and when I looked at the `SupportsDebugging` function it looks like it looks for `project`. https://github.com/microsoft/aspire/blob/45bbcb799bc647bd5ca376b9a8376a65efc2a1dc/src/Aspire.Hosting/Utils/ExtensionUtils.cs#L22

dscpinheiro (approved):

peterrsongg (approved):

muhammad-othman (approved):

normj:
Put in draft mode till we confirm some findings from the Aspire team

adamint:
Please see https://github.com/microsoft/aspire/pull/15714

normj:
Closing because there is a better fix coming from the Aspire team that avoids adding calls to experimental APIs.

PR #172, opened 2026-04-02
fix build

GarrettBeatty:
workaround to fix ci

### DIFF
--- a/buildtools/ci.buildspec.yml
+++ b/buildtools/ci.buildspec.yml
@@ -10,6 +10,7 @@ phases:
       - cdk --version      
   build:
     commands:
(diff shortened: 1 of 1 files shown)


philasmar (changes requested):
I am not a fan of this approach. I'd recommend you follow the same approach as the lambda repo https://github.com/aws/aws-lambda-dotnet/blob/master/buildtools/ci.buildspec.yml#L9. We delete CodeBuild's global.json and then we install the versions of .NET we care about.

philasmar (approved):

dscpinheiro (approved):

PR #173, opened 2026-04-03
Support WithEnvironment on compute publish targets (#169)

boblodgett:
## Description

Resolve `WithEnvironment` callbacks during publish for all AWS compute targets. Previously, `EnvironmentCallbackAnnotation` entries (added by `.WithEnvironment()` calls) were silently dropped during publish — only `WithReference`-derived env vars were processed.

This PR adds three capabilities to the publish pipeline:

1. **EnvironmentCallbackAnnotation resolution** — Executes all `WithEnvironment` callbacks in publish mode and merges the resulting env vars into CDK construct properties. `WithEnvironment` values take precedence over `WithReference`-derived values.

2. **IValueProvider resolution** — Handles non-string env var values (e.g., `ParameterResource`, `ReferenceExpression`) by calling `GetValueAsync()` to resolve them at publish time. Values that resolve to null or throw exceptions are logged as warnings and skipped gracefully.

3. **Lambda 4KB env var limit warning** — Calculates total UTF-8 byte size of Lambda function environment variables and logs an error if exceeding 4096 bytes or a warning at 80% of the limit.

The core fix lives in `AbstractAWSPublishTarget.ProcessRelationShipsAsync()` (base class), so all 4 publish targets benefit: ECS Fargate, ECS Fargate Express, ECS Fargate with ALB, and Lambda.

## Motivation and Context

Fixes #169. Users calling `.WithEnvironment("KEY", "value")` on `ProjectResource` or `ContainerResource` expected those env vars to be present when deployed to AWS. Instead, they were silently dropped because the publish pipeline only processed `WithReference`-based env vars.

## Testing

- Added 9 new unit tests in `ProcessRelationShipsTests.cs` covering:
  - Static string values resolved and added
  - Multiple callbacks all invoked
  - `WithEnvironment` precedence over existing (WithReference-derived) env vars
  - `IValueProvider` values resolved (simulates ParameterResource)
  - Null-resolving `IValueProvider` skipped with warning
  - Exception-throwing `IValueProvider` handled gracefully
  - Unsupported value types skipped with warning
  - No-callback resources unaffected
  - Callback context uses Publish operation mode
- All 88 unit tests pass (0 failures)

## Breaking Changes Assessment

No breaking changes. The `ProcessRelationShips` → `ProcessRelationShipsAsync` method rename is behind the `[Experimental(ASPIREAWSPUBLISHERS001)]` attribute, so it is not part of the stable public API surface. The behavior change is additive — env vars that were previously dropped are now resolved.

[...]

PR #193, opened 2026-05-08
Support WithEnvironment and ParameterResource on publish targets

### LINKED ISSUE #169: [Deployment] Support WithEnvironment on compute publish targets (ECS Fargate, ECS Fargate Express, Lambda)
### Describe the feature

`WithEnvironment` calls on `ProjectResource` and `ContainerResource` in the AppHost are silently dropped when publishing to AWS. The deployment pipeline processes `WithReference`-based environment variables but never reads `EnvironmentCallbackAnnotation`, which is the annotation `WithEnvironment` creates on the resource. This means any explicit environment configuration set via the standard Aspire API has no effect on the deployed AWS resources.
 
Similarly, `ParameterResource` values — the Aspire-native way to inject secrets and configurable values — are not resolved or propagated to AWS compute targets during publish. Parameters defined via `builder.AddParameter()` and passed through `WithEnvironment` or `WithReference` should resolve during the publish pipeline, just as they do for Docker and Azure targets.
 
This affects all three compute publish targets: ECS Fargate, ECS Fargate Express, and Lambda.

### Use Case

When building Aspire applications that target AWS, I need to set environment variables on my compute resources — feature flags, service URLs for non-Aspire dependencies, tuning parameters, and secrets managed as parameters. The idiomatic Aspire way to do this is `WithEnvironment` and `AddParameter`:
 
```csharp
var spaClientKey = builder.AddParameter("spa-client-key", secret: true);
 
var api = builder.AddProject<Projects.Api>("api")
    .WithEnvironment("FEATURE_NEW_WORKFLOW", "true")
    .WithEnvironment("EXTERNAL_PAYMENT_URL", "https://pay.example.com")
    .WithEnvironment("SPA_CLIENT_KEY", spaClientKey);
```
 
This works when publishing to Docker and Azure targets, but the variables and parameter values are silently lost when publishing to AWS via `aspire publish` or `aspire deploy`. There's no error or warning — the env vars just don't show up on the ECS task definition or Lambda function configuration. This breaks the Aspire promise of a consistent dev-to-cloud inner loop and forces AWS-specific workarounds (e.g., CDK construct callbacks to manually inject env vars, or wiring SSM/Secrets Manager references by hand).

### Proposed Solution

During the publish pipeline, resolve `EnvironmentCallbackAnnotation` entries and `ParameterResource` references on compute resources and merge them into the target's `EnvironmentVariables` dictionary before CDK construct generation. The plumbing is mostly there — each compute publish target (`ECSFargateServicePublishTarget`, `ECSFargateServiceWithALBPublishTarget`, `ECSFargateServiceExpressPublishTarget`, `LambdaFunctionPublishTarget`) already exposes an `EnvironmentVariables` property that maps to the CDK construct's `Environment` dictionary.
 
The key integration point would be in `AbstractAWSPublishTarget` (or in each target's `GenerateConstructAsync`), roughly:
 
```csharp
// Resolve WithEnvironment annotations
var envCallbacks = resource.Annotations.OfType<EnvironmentCallbackAnnotation>();
// Execute callbacks, resolve ParameterResource references, and merge into EnvironmentVariables
```
 
**ParameterResource should become CloudFormation parameters.** Rather than baking parameter values into the template as literals, `ParameterResource` instances should map to first-class CloudFormation template parameters (via `CfnParameter` in CDK). This keeps the synthesized template reusable across environments and enables value injection at deploy time via `cdk deploy --parameters`, CI/CD pipeline configuration, or the CloudFormation console. The container/function environment variable would then reference the CloudFormation parameter using `Fn::Ref` rather than a hardcoded string. For `secret: true` parameters, the CloudFormation parameter should be declared with `NoEcho: true` and ideally backed by an SSM SecureString or Secrets Manager reference (future).

Additional considerations:
- **Static `WithEnvironment` values** (plain strings, not parameter references) can be inlined directly into the task definition or function configuration — no CloudFormation parameter needed.
- **Callback environment variables** may reference other Aspire resources via expressions. The publish pipeline would need to resolve these the same way `WithReference` env vars are resolved today.
- **Ordering** — explicit `WithEnvironment` values should probably take precedence over (or at minimum not silently overwrite) reference-derived env vars from `WithReference`.
- **Lambda** has a 4 KB limit on environment variable payloads — a warning during publish would be a nice touch if the resolved set approaches that limit.

### Other Information

The `WithReference` path already resolves environment variables correctly for all three compute targets, so this is a gap in coverage rather than a net-new capability. The Docker and Azure publish integrations in the core Aspire repo handle both `WithEnvironment` and `ParameterResource` during their publish pipelines, so there's prior art to reference for the resolution approach.
 
The CloudFormation parameter mapping is particularly important for CI/CD workflows — teams deploying the same template across dev/staging/prod need to vary parameters without re-synthesizing.
 
Relevant source files:
- `src/Aspire.Hosting.AWS/Deployment/CDKPublishTargets/AbstractAWSPublishTarget.cs`
- `src/Aspire.Hosting.AWS/Deployment/CDKPublishTargets/ECSFargateServicePublishTarget.cs`
- `src/Aspire.Hosting.AWS/Deployment/CDKPublishTargets/ECSFargateServiceWithALBPublishTarget.cs`
- `src/Aspire.Hosting.AWS/Deployment/CDKPublishTargets/ECSFargateServiceExpressPublishTarget.cs`
- `src/Aspire.Hosting.AWS/Deployment/CDKPublishTargets/LambdaFunctionPublishTarget.cs`

### Acknowledgements

- [x] I may be able to implement this feature request

Aspire.Hosting.AWS 13.0.0

### Targeted .NET Platform

.NET 10

macOS Tahoe

GarrettBeatty:
## Summary
- Resolves `WithEnvironment` callbacks during publish for all AWS compute targets (ECS Fargate, ECS Fargate Express, ECS Fargate with ALB, Lambda)
- Maps `ParameterResource` values to CloudFormation parameters (resolved at deploy time, not publish time), avoiding the hanging issue found in #173
- Reference-derived env vars (from `WithReference`) take precedence over `WithEnvironment` on key collision since they contain CDK token values (Fn::GetAtt, etc.)

## Context
This is a rework of #173 incorporating the approach from @jasonmueller's [branch](https://github.com/jasonmueller/integrations-on-dotnet-aspire-for-aws/tree/f/support-withenvironment). The key insight is that `ParameterResource` should NOT be resolved at publish time — it should become a CloudFormation parameter that gets its value at deploy time. This avoids the deployment hang that @normj identified when `GetValueAsync()` was called on value providers that aren't resolvable at publish time.

### Key design decisions:
1. **ParameterResource → CfnParameter**: Each `ParameterResource` creates a CFN parameter on the stack; the env var value becomes a `Ref` token. Secret parameters get `NoEcho: true`.
2. **CfnParameter caching**: Parameters are cached per name so multiple resources sharing the same parameter only create one CFN parameter on the stack.
3. **Callback error handling**: Failed callbacks (e.g., reference-injected ones that need endpoint allocation) are caught and logged at Trace level, since those env vars are handled separately by `GetReferenceConnectionInfo`.
4. **Precedence**: `WithEnvironment` values are applied first, then `WithReference`-derived values overwrite on collision (CDK tokens are the authoritative infrastructure values).

Fixes #169

## Test plan
- [x] All 79 existing unit tests pass
- [x] New integration tests: `TestPublishWithEnvironment`, `TestPublishWithEnvironmentParameter`, `TestPublishWithEnvironmentAndReference`
- [x] Verify ParameterResource creates CFN Parameter with correct `Ref` in generated template
- [x] Verify static string env vars appear in all compute target templates
- [x] Verify combined WithEnvironment + WithReference both appear correctly

image

image
image

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.autover/changes/26f7825a-0b8c-4ffb-94a8-7127fba427c3.json
+++ b/.autover/changes/26f7825a-0b8c-4ffb-94a8-7127fba427c3.json
@@ -1,0 +1,12 @@
+{
+  "Projects": [
+    {
+      "Name": "Aspire.Hosting.AWS",
+      "Type": "Patch",
+      "ChangelogMessages": [
+        "Resolve WithEnvironment callbacks during publish for all compute targets (ECS Fargate, ECS Fargate Express, ECS Fargate with ALB, Lambda)",
+        "Support ParameterResource in WithEnvironment by mapping to CloudFormation parameters at deploy time"
+      ]
+    }
+  ]
+}
--- a/playground/Publishing/Publishing.AppHost/Program.cs
+++ b/playground/Publishing/Publishing.AppHost/Program.cs
@@ -30,8 +30,15 @@ var localDevQueue = cdkStackResource.AddSQSQueue("LocalDevQueue");
 
 var cache = builder.AddValkey("cache");
 
+// ParameterResource examples — these become CloudFormation parameters at deploy time
+builder.Configuration["Parameters:api-key"] = "default-api-key";
+builder.Configuration["Parameters:db-connection"] = "Server=localhost;Database=mydb";
+var apiKey = builder.AddParameter("api-key");
+var dbConnection = builder.AddParameter("db-connection", secret: true);
+
 var frontend = builder.AddProject<Projects.Frontend>("Frontend")
         .WithEnvironment("ENV_LAMBDA_1", "LambdaValue1")
+        .WithEnvironment("API_KEY", apiKey)
         .WithEnvironment(callback: (env) =>
         {
             env.EnvironmentVariables.Add("ENV_LAMBDA_2", "LambdaValue2");
@@ -43,6 +50,7 @@ var frontend = builder.AddProject<Projects.Frontend>("Frontend")
 
 builder.AddProject<Projects.Backend>("backend")
         .WithEnvironment("ENV_LAMBDA_1", "LambdaValue1")
+        .WithEnvironment("DB_CONNECTION", dbConnection)
         .WithEnvironment(callback: (env) =>
         {
             env.EnvironmentVariables.Add("ENV_LAMBDA_2", "LambdaValue2");
@@ -64,6 +72,7 @@ builder.AddAWSLambdaFunction<Projects.SQSProcessorFunction>("SQSProcessorFunctio
             }
         })
         .WithEnvironment("ENV_LAMBDA_1", "LambdaValue1")
+        .WithEnvironment("API_KEY", apiKey)
         .WithEnvironment(callback: (env) =>
         {
             env.EnvironmentVariables.Add("ENV_LAMBDA_2", "LambdaValue2");
--- a/playground/Publishing/Publishing.AppHost/Program.cs
+++ b/playground/Publishing/Publishing.AppHost/Program.cs
@@ -31,12 +31,22 @@ var localDevQueue = cdkStackResource.AddSQSQueue("LocalDevQueue");
 var cache = builder.AddValkey("cache");
 
 var frontend = builder.AddProject<Projects.Frontend>("Frontend")
+        .WithEnvironment("ENV_LAMBDA_1", "LambdaValue1")
+        .WithEnvironment(callback: (env) =>
+        {
+            env.EnvironmentVariables.Add("ENV_LAMBDA_2", "LambdaValue2");
+        })
         .WithExternalHttpEndpoints()
         .WithReference(cache)
         .WaitFor(cache);
 
 
 builder.AddProject<Projects.Backend>("backend")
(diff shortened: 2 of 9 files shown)


philasmar (approved):

normj (approved):
Thanks @jasonmueller for the approach. We should also extend the deploy step to resolve the parameters and pass those that were mapped as CF parameters into the "cdk deploy" command. With this PR you can publish but then have to directly use the CDK to get the parameters applied. I don't want to hold up releasing this with the deploy step to unblock @jasonmueller.

normj:
PR has been released as part of version `13.1.0` of Aspire.Hosting.AWS.

PR #197, opened 2026-05-13
ci: add contents: read to change-file-in-pr workflow

arpitjain099:
Single-workflow PR. `change-file-in-pr.yml` runs on `pull_request`, checks out the PR, and uses `tj-actions/changed-files` to flag whether the PR includes a release-note change. Pure read; no comments, pushes, or release writes. `contents: read` is the minimum.


GarrettBeatty:
image
 we havent had issues with this workflow running so not sure if this is needed?

PR #199, opened 2026-05-14
Add support for custom container runtime (e.g. podman) in CDK publish

halllo:
# Support Podman (and other OCI runtimes) in CDK publishing steps

## Description

Read `ASPIRE_CONTAINER_RUNTIME` from the environment (defaulting to `docker`) wherever the CDK publish pipeline previously hardcoded the `docker` binary:

- `DefaultTarballContainerImageBuilder`: uses the runtime when building the `save` command
- `CDKPublishingStep.FixCDKAssetsFileForWindows`: uses the runtime when building the PowerShell `load` command

## Motivation and Context

Users running Podman instead of Docker could not publish without symlinking `podman` to `docker`, because both publish paths hardcoded the `docker` binary name. The workaround is unsupported and breaks in environments where both runtimes coexist or where the user cannot write to system directories.

`ASPIRE_CONTAINER_RUNTIME` is already used by the core Aspire CLI to select between Docker and Podman for local run scenarios. Using the same variable keeps behaviour consistent across the Aspire ecosystem and requires no new user-facing API.

Usage with Podman:

```bash
ASPIRE_CONTAINER_RUNTIME=podman aspire publish -o ./publish
```

When the env var is unset the behaviour is identical to before (`docker`).

## Testing

- Verified the project builds without errors or warnings
- Manually tested `aspire publish` with `ASPIRE_CONTAINER_RUNTIME=podman` on macOS with Podman installed
- Manually tested `aspire publish` without the env var set to confirm the default `docker` path is unchanged

No breaking changes. When `ASPIRE_CONTAINER_RUNTIME` is unset the pipeline behaves exactly as before.

N/A

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed

## License

- [x] I confirm that this pull request can be released under the [Apache 2 license](http://aws.amazon.com/apache2.0/)

### DIFF
--- a/src/Aspire.Hosting.AWS/Deployment/Services/DefaultTarballContainerImageBuilder.cs
+++ b/src/Aspire.Hosting.AWS/Deployment/Services/DefaultTarballContainerImageBuilder.cs
@@ -1,41 +1,27 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 
-using System.Runtime.InteropServices;
+#pragma warning disable ASPIRECONTAINERRUNTIME001
+
 using Aspire.Hosting.ApplicationModel;
 using Aspire.Hosting.AWS.Utils.Internal;
+using Aspire.Hosting.Publishing;
 using Microsoft.Extensions.Logging;
 
 namespace Aspire.Hosting.AWS.Deployment.Services;
 
-internal class DefaultTarballContainerImageBuilder(ILogger<DefaultTarballContainerImageBuilder> logger, IProcessCommandService processCommandService) : ITarballContainerImageBuilder
+internal class DefaultTarballContainerImageBuilder(ILogger<DefaultTarballContainerImageBuilder> logger, IProcessCommandService processCommandService, IContainerRuntime containerRuntime) : ITarballContainerImageBuilder
 {
     public async Task<string> CreateTarballImageAsync(ProjectResource resource, CancellationToken cancellationToken)
     {
         var tarballFilePath = Path.GetTempFileName() + ".tar";
 
         var imageTag = resource.Name.ToLower() + ":latest";
-        var runtime = Environment.GetEnvironmentVariable("ASPIRE_CONTAINER_RUNTIME") ?? "docker";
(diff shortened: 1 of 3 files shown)


normj (approved):

src/Aspire.Hosting.AWS/Deployment/Services/DefaultTarballContainerImageBuilder.cs
```diff
@@ -14,7 +14,8 @@ public async Task<string> CreateTarballImageAsync(ProjectResource resource, Canc
         var tarballFilePath = Path.GetTempFileName() + ".tar";
 
         var imageTag = resource.Name.ToLower() + ":latest";
-        var dockerSaveCommand = $"docker save -o {tarballFilePath} {imageTag}";
+        var runtime = Environment.GetEnvironmentVariable("ASPIRE_CONTAINER_RUNTIME") ?? "docker";
```
eerhardt: Can this use `IContainerRuntime` instead? should be able to use

`services.GetRequiredService<IContainerRuntime>();`

halllo: Yes, `IContainerRuntimeResolver`.

normj: @eerhardt How concerned should I be about using the `IContainerRuntimeResolver` given it is marked as experimental? Granted it is being used in APIs in this package that are marked as experimental as well but if you know that API is likely to change or be removed I would rather not base logic on it.

normj (commented):
Waiting on @eerhardt response to my question about `IContainerRuntimeResolver` before approving

src/Aspire.Hosting.AWS/Deployment/Services/DefaultTarballContainerImageBuilder.cs
normj: @eerhardt I saw 13.4 was released today and `IContainerRuntimeResolver` is still marked as experimental. Any update on its status?

PR #200, opened 2026-05-14
Add VPC endpoint creation for ECS Express services in public subnets

halllo:
Add VPC Interface Endpoints for ECR and a Gateway Endpoint for S3 to the default CDK stack whenever an ECS Express service is configured. Also fixes a CloudFormation race condition that causes a "mixed subnet type" deployment error.

A new method `EnsureECSExpressVpcEndpoints()` is called from `ApplyCfnExpressGatewayServiceDefaults` and creates three VPC endpoints:

| Endpoint | Type | Purpose |
|---|---|---|
| `ecr.api` | Interface | ECR control plane: authentication, manifest fetch |
| `ecr.dkr` | Interface | ECR data plane: image layer download |
| `s3` | Gateway | ECR stores image layers in S3; free and route-table based |

A boolean guard ensures the endpoints are created only once even when multiple ECS Express services are deployed in the same stack.

An explicit `DependsOn` on `vpc.InternetConnectivityEstablished` is added to the `CfnExpressGatewayService` construct to prevent a race condition in CloudFormation (see second failure mode below).

### Failure mode 1 — `ResourceInitializationError` at task startup

`PublishAsECSFargateExpressService` fails at task startup with a `ResourceInitializationError` on every deployment. The root cause is a combination of two AWS design decisions that interact badly:

1. `ExpressGatewayServiceNetworkConfiguration` does not expose `assignPublicIp` — AWS removed it from the API. Tasks always run with a private IP only.
2. The default VPC places tasks in **public subnets** (so the managed ALB is internet-facing), but a resource in a public subnet with no public IP has no outbound internet access — the Internet Gateway only forwards packets for resources that have a public IP, and there is no NAT Gateway.

```
                    INTERNET
                       │
                       ▼
              ┌────────────────┐
              │  ALB (managed  │  ← AWS creates this internally
              │  by ECS        │    (hidden inside ECS Express)
              │  Express)      │
              └────────┬───────┘
                       │
          ╔════════════╧═══════════════╗
          ║     PUBLIC SUBNET          ║
          ║                            ║
          ║  ┌──────────────────────┐  ║
          ║  │  ECS Task            │  ║  ← Private IP only!
          ║  │  IP: 10.0.1.42 only  │  ║    assignPublicIp not
          ║  │  (no public IP)      │  ║    supported by API
          ║  └──────────┬───────────┘  ║
          ║             │              ║
          ╚═════════════╪══════════════╝
                        │ wants to pull image from ECR
                        ▼
              ┌──────────────────┐
              │  ECR             │  ✗ Unreachable
              │  (public internet│    No public IP → IGW blocks it
              │   endpoint)      │    No NAT → no alternative path
              └──────────────────┘

  Result: ResourceInitializationError: unable to pull secrets or
          registry auth: ... i/o timeout
```

ECS Express is designed to be used with AWS PrivateLink. Adding VPC endpoints routes ECR traffic over the private AWS backbone without touching the public internet, which is the intended architecture:

```
          ╔════════════════════════════════════╗
          ║     PUBLIC SUBNET                  ║
          ║                                    ║
          ║  ┌──────────────────────┐          ║
          ║  │  ECS Task            │          ║
          ║  │  IP: 10.0.1.42 only  │          ║
          ║  └──────────┬───────────┘          ║
          ║             │ to ECR               ║
          ║             ▼                      ║
          ║  ┌──────────────────────────────┐  ║
          ║  │  VPC Interface Endpoints     │  ║
          ║  │  com.amazonaws.{region}      │  ║
          ║  │    .ecr.api    (control plane)│  ║
          ║  │    .ecr.dkr    (data plane)  │  ║
          ║  │    .s3 gateway (layer store) │  ║
          ║  └──────────┬───────────────────┘  ║
          ╚═════════════╪══════════════════════╝
                        │ AWS PrivateLink
                        │ (never touches internet)
                        ▼
              ┌──────────────────┐
              │  ECR             │  ✓ Reachable
              └──────────────────┘
```

`PrivateDnsEnabled = true` on the Interface Endpoints causes DNS within the VPC to resolve ECR hostnames to the private endpoint IP rather than the public internet IP — no application code changes are needed.

**Cost note:** The two Interface Endpoints (`ecr.api`, `ecr.dkr`) incur standard AWS PrivateLink hourly and data-processing charges. The S3 Gateway Endpoint is free. This is significantly cheaper than the NAT Gateway required by the `PublishAsECSFargateServiceWithALB` workaround.

### Failure mode 2 — `InvalidRequest` during CloudFormation deployment

Even before tasks start, the stack can fail during `CREATE_IN_PROGRESS` with:

```
Invalid request provided: Provided subnet list contains both public and private subnets.
Only one type of subnet is allowed.
```

**Root cause — CloudFormation race condition.** Subnet IDs are passed to `CfnExpressGatewayService` as string tokens, which create an implicit CloudFormation dependency on the *subnet resource* only — not on the IGW route entries that classify a subnet as "public". CloudFormation creates the ECS Express service and the public subnet IGW routes in parallel. If ECS Express validates subnet types before both IGW routes exist, one subnet appears public (route ready) and the other appears private (route not yet ready), producing a mixed-type list.

**Fix:** `fargateService.Node.AddDependency(vpc.InternetConnectivityEstablished)` adds an explicit `DependsOn` on all IGW route resources, ensuring CloudFormation fully establishes internet connectivity for the public subnets before the ECS Express service is created.

## Testing

Four new unit tests were added to `ApplyDefaultsTests`:

- **`ApplyCfnExpressGatewayServiceDefaults_CreatesVpcEndpoints`** — asserts that all three VPC endpoint constructs (`ECSExpressEcrApiEndpoint`, `ECSExpressEcrDkrEndpoint`, `ECSExpressS3Endpoint`) are present in the CDK stack after a single call to `ApplyCfnExpressGatewayServiceDefaults`.

- **`ApplyCfnExpressGatewayServiceDefaults_CreatesVpcEndpointsOnlyOnce`** — calls `ApplyCfnExpressGatewayServiceDefaults` twice to simulate two ECS Express services in the same stack, and asserts each endpoint node appears exactly once.

- **`ApplyCfnExpressGatewayServiceDefaults_DoesNotCreateVpcEndpoints_ByDefault`** — asserts that endpoints are NOT created when no config is passed (`EnableVpcEndpoints` defaults to `false`).

- **`ApplyCfnExpressGatewayServiceDefaults_DoesNotCreateVpcEndpoints_WhenCustomNetworkConfigurationProvided`** — asserts that endpoints are NOT created when the caller supplies a custom `NetworkConfiguration`.

All 92 unit tests pass.

No breaking changes. `EnsureECSExpressVpcEndpoints` is `protected virtual`, allowing subclasses to override endpoint creation. The boolean guard ensures idempotency. No existing public API surface or behavior is altered — the only observable difference is that stacks using `PublishAsECSFargateExpressService` will include three additional VPC endpoint resources when opted in, and the `CfnExpressGatewayService` construct will have explicit `DependsOn` entries for the VPC's public subnet routes.

N/A

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

## License

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 *.userosscache
 *.sln.docstates
 *.DS_Store
+*.lscache
 
 playground/**/launchSettings.json
 **/.aspire/**
--- a/src/Aspire.Hosting.AWS/Deployment/CDKDefaultAttributes.cs
+++ b/src/Aspire.Hosting.AWS/Deployment/CDKDefaultAttributes.cs
@@ -84,3 +84,33 @@ public class DefaultElastiCacheServerlessSecurityGroupAttribute : Attribute
 {
 
 }
+
+/// <summary>
+/// Attribute applied on a property or field in the CDK deployment stack of type <see cref="Amazon.CDK.AWS.EC2.InterfaceVpcEndpoint"/>.
+/// This will be used as the default ECR API interface VPC endpoint for ECS Express services, allowing image pulls via PrivateLink.
+/// </summary>
+[AttributeUsage(AttributeTargets.Property | AttributeTargets.Field)]
+public class DefaultECSExpressEcrApiEndpointAttribute : Attribute
+{
+
+}
+
+/// <summary>
+/// Attribute applied on a property or field in the CDK deployment stack of type <see cref="Amazon.CDK.AWS.EC2.InterfaceVpcEndpoint"/>.
+/// This will be used as the default ECR Docker (DKR) interface VPC endpoint for ECS Express services, allowing image layer pulls via PrivateLink.
+/// </summary>
+[AttributeUsage(AttributeTargets.Property | AttributeTargets.Field)]
+public class DefaultECSExpressEcrDkrEndpointAttribute : Attribute
+{
+
+}
+
+/// <summary>
+/// Attribute applied on a property or field in the CDK deployment stack of type <see cref="Amazon.CDK.AWS.EC2.GatewayVpcEndpoint"/>.
+/// This will be used as the default S3 gateway VPC endpoint for ECS Express services, routing ECR image layer traffic through S3 PrivateLink.
+/// </summary>
+[AttributeUsage(AttributeTargets.Property | AttributeTargets.Field)]
+public class DefaultECSExpressS3EndpointAttribute : Attribute
+{
+
+}
--- a/src/Aspire.Hosting.AWS/Deployment/CDKDefaultAttributes.cs
+++ b/src/Aspire.Hosting.AWS/Deployment/CDKDefaultAttributes.cs
@@ -114,3 +114,13 @@ public class DefaultECSExpressS3EndpointAttribute : Attribute
 {
 
 }
+
+/// <summary>
+/// Attribute applied on a property or field in the CDK deployment stack of type <see cref="Amazon.CDK.AWS.EC2.InterfaceVpcEndpoint"/>.
+/// This will be used as the default CloudWatch Logs interface VPC endpoint for ECS Express services, allowing the awslogs log driver to reach CloudWatch Logs via PrivateLink.
+/// </summary>
+[AttributeUsage(AttributeTargets.Property | AttributeTargets.Field)]
+public class DefaultECSExpressLogsEndpointAttribute : Attribute
+{
+
+}
--- a/src/Aspire.Hosting.AWS/Deployment/CDKDefaults/CDKDefaultsProvider.DefaultConstructs.cs
+++ b/src/Aspire.Hosting.AWS/Deployment/CDKDefaults/CDKDefaultsProvider.DefaultConstructs.cs
@@ -339,6 +339,39 @@ public partial class CDKDefaultsProvider
         });
     }
 
+    private InterfaceVpcEndpoint? _defaultECSExpressLogsEndpoint;
+    public InterfaceVpcEndpoint GetDefaultECSExpressLogsEndpoint()
+    {
+        if (_defaultECSExpressLogsEndpoint == null)
+        {
+            var definedDefault = FindDefaultConstructByAttribute<DefaultECSExpressLogsEndpointAttribute, InterfaceVpcEndpoint>();
+            if (definedDefault != null)
+            {
+                _defaultECSExpressLogsEndpoint = definedDefault;
+            }
+            else
+            {
+                _defaultECSExpressLogsEndpoint = CreateDefaultECSExpressLogsEndpoint();
+            }
+        }
+        return _defaultECSExpressLogsEndpoint;
+    }
+
+    protected virtual InterfaceVpcEndpoint CreateDefaultECSExpressLogsEndpoint()
+    {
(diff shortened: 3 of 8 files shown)


normj:
I'm thinking the current version is working for me because the subnets in my default VPC have "Auto-assign public IPv4 address" set to true and for you that is not set. Since this adds additional cost and isn't always needed I think enabling the gateways needs to be an opt-in mechanism.

normj (commented):
The PR looks good. I'm trying to reproduce your initial failure in my account and then try enabling the endpoints.

normj:
@halllo I created a new VPC with only public subnets and "Auto-assign public IPv4 address" turned off but in my testing when I deploy to that VPC Express still makes sure the tasks get a IPv4 address and deployment works. Can you describe your VPC setup and how the subnets are public through the internet gateway?

halllo:
@normj My VPC is an AWS Control Tower-provisioned VPC — there is no default AWS VPC in this account at all. It has three private subnets only (MapPublicIpOnLaunch: false, tagged Network: Private), no Internet Gateway, and no NAT Gateway. There is no route to the public internet whatsoever.

So the difference from your test is not just "Auto-assign public IPv4 off" — I have no public subnets and no IGW. Even if ECS Express were to assign a public IP, there's no gateway to route through. VPC endpoints are the only viable connectivity path to ECR in this setup.

This is a common setup in enterprise AWS accounts managed by Control Tower. Would it make sense to scope this PR to explicitly target this scenario (private-only VPCs / no IGW), while keeping the default behavior unchanged for accounts with a standard default VPC?

normj:
Ah that helps give context so you are making an internal facing ECS express and our logic is assuming it is defaulting for public. In fact the default logic for selecting the subnets is to grab public subnets https://github.com/aws/integrations-on-dotnet-aspire-for-aws/blob/main/src/Aspire.Hosting.AWS/Deployment/CDKDefaults/CDKDefaultsProvider.ECSFargateExpressService.cs#L71

With this change you would have to enable the endpoints and provide your own `NetworkConfiguration` on the props callback.

I wonder if we should generalize this change more and instead of having the `EnableVpcEndpoints` property we have an `InternetFacing` property defaulted to `true`. If set to `false` we add the endpoint you PR has code for and change the logic in `ApplyCfnExpressGatewayServiceDefaults` to grab private subnets. What do you think?

I appreciate your patience helping me understand the issue.

halllo:
When I deploy the stack in my account, I do want the ECS express to be publicly reachable through the ALB.

normj:
If there are no public subnets then how would the ALB be be public? My understanding if the subnets set on the network configuration are private then ECS express will make the ALB it creates be an "internal" ALB.

halllo:
Yes, from my understanding deploying the stack creates a new public subnet (so the managed ALB is internet-facing), but the resource in the public subnet does not get a public IP. It does not need it, because it gets the traffic from the ALB. It only needs to be able to reach out to ECR.

PR #203, opened 2026-05-14
Add support for S3 static websites with CloudFront and API

halllo:
Adds a new AWS CDK publish target, `PublishAsS3WithCloudFront`, that deploys Aspire JavaScript resources (`AddViteApp`, `AddNpmApp`, etc.) as static websites hosted on a private Amazon S3 bucket fronted by a CloudFront distribution for HTTPS and global CDN delivery.

**New publish target: `S3StaticWebsitePublishTarget`**

- Runs the app's build script (`npm run build`, or the package manager / script configured via Aspire annotations) before synthesizing CDK constructs, injecting any `WithReference`-derived environment variables into the build process. Note: for AWS-published backends (e.g. ECS Fargate), these values are CloudFormation tokens resolved at deploy time — use relative API paths and `WithCloudFrontBackendBehavior` for frontend-to-backend calls rather than embedding the injected URL in the static build
- Creates a private S3 bucket (public access blocked, SSL enforced) accessed via CloudFront Origin Access Control
- Deploys built assets with `BucketDeployment` and invalidates the CloudFront cache on every deploy
- Emits a CloudFormation output with the CloudFront URL (`{name}-CloudFrontUrl`)
- Registers `JavaScriptAppResource` as the default auto-matched target, so JavaScript apps deploy to S3+CloudFront automatically without explicit configuration
- Validates `OutputPath` at synthesis time: rooted paths and paths that escape the working directory are rejected to prevent accidental deployment of unintended local files

**New extension methods**

```csharp
var backend = builder.AddProject<Projects.Backend>("backend")
    .PublishAsECSFargateServiceWithALB();

var frontend = builder.AddViteApp("frontend", "../frontend")
    .WithReference(backend)
    .WithCloudFrontBackendBehavior("/todos/*", backend)  // route API paths to backend
    .PublishAsS3WithCloudFront(config =>
    {
        config.OutputPath = "dist/browser"; // default: "dist"
    });
```

**CloudFront backend behaviors (`WithCloudFrontBackendBehavior`)**

Allows specific path patterns (e.g. `/todos/*`) to be forwarded to a backend origin instead of being served from S3 or falling through to the SPA index-fallback. The backend's protocol (HTTP or HTTPS) is detected automatically from its service-discovery URL. Caching is disabled on backend behaviors and all HTTP methods are forwarded. When a pattern ends with `/*`, the bare path (e.g. `/todos`) is automatically registered as a second exact-match behavior — CloudFront's `/*` glob does not match the path without a trailing slash.

**Full CDK customization via callbacks** (consistent with other publish targets):

- `PropsBucketCallback` / `ConstructBucketCallback`
- `PropsBucketDeploymentCallback`
- `PropsDistributionCallback` / `ConstructDistributionCallback`

**`IStaticSiteBuilder` / `DefaultStaticSiteBuilder`**

Extracted the build-script invocation into a dedicated, mockable service so it can be unit-tested independently of CDK construct creation. `BuildAsync` checks the `CancellationToken` before starting the build so that a cancelled deploy does not launch a new build process.

**Working-directory capture**

Aspire's `PublishAsStaticWebsite()` (YARP dev proxy) transforms the resource type from `JavaScriptAppResource` to `ExecutableContainerResource` in the publish pipeline. To survive this transformation, the working directory is captured into `PublishS3WithCloudFrontAnnotation` at registration time rather than being read from the resource at CDK-synthesis time.

## Motivation and Context

There was no first-class path to deploy a Vite/Angular/React app from an Aspire AppHost to AWS. Users had to hand-author CDK stacks or use separate tooling outside of Aspire's publish/deploy flow. This PR closes that gap with the same `PublishAs*` ergonomics used by the ECS Fargate and Lambda targets.

The CloudFront backend-behavior feature was motivated by the common SPA + backend API pattern: without it, the SPA 404→index.html error response intercepts all unrecognized paths including API prefixes, so API calls never reach the backend.

## Testing

**Unit tests** — 21 tests across two new test classes:

- `DefaultStaticSiteBuilderTests` (6 tests): default npm command, custom package manager annotation, custom build script annotation, build args with `--`, environment variable pass-through, non-zero exit code throws.
- `S3StaticWebsitePublishTargetTests` (15 tests): `IsDefaultPublishTargetMatch` for JS and non-JS resources; bucket defaults (private, SSL enforced); distribution creation; CloudFront URL output; SPA error responses (403/404 → 200/index.html asserted on props); all five CDK callbacks; `IStaticSiteBuilder` invocation with env vars; `GetReferenceConnectionInfo` returns HTTPS endpoint; custom `OutputPath` (isolated working directory proves the path is actually used); backend behavior missing-annotation error; backend behavior unsupported construct type error.

All 21 tests pass. All pre-existing tests continue to pass.

**End-to-end** — Deployed a real Angular frontend with a backend ECS Fargate service using `WithCloudFrontBackendBehavior("/todos/*", backend)`. Verified via `curl` and the AWS CLI that:

- Static assets are served from CloudFront/S3
- `/todos/*` and the bare `/todos` path both reach the backend (JSON response, HTTP 200)
- SPA client-side routes return `index.html` as expected

No breaking changes. All new public API is additive and gated behind the existing `[Experimental(ASPIREAWSPUBLISHERS001)]` attribute. The `IStaticSiteBuilder` interface is new and `public`, but it was not present in any previous release.

N/A — infrastructure-only change with no UI.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)

## Checklist
- [x] My change requires a change to the documentation
- [x] I have updated the documentation accordingly

## License

[issues]: https://github.com/aws/aws-sdk-net/issues
[license]: http://aws.amazon.com/apache2.0/
[cla]: http://en.wikipedia.org/wiki/Contributor_License_Agreement


src/Aspire.Hosting.AWS/Deployment/CDKPublishTargets/S3StaticWebsitePublishTarget.cs
```diff
@@ -0,0 +1,216 @@
...
+        var buildEnvVars = connectionPoints.EnvironmentVariables ?? new Dictionary<string, string>();
+
+        await siteBuilder.BuildAsync(resource, workingDirectory, buildEnvVars, cancellationToken);
+
+        var buildOutputPath = Path.GetFullPath(Path.Combine(workingDirectory, config.OutputPath));
+
+        var context = CreatePublishTargetContext(environment);
+
+        // --- S3 Bucket ---
+        var bucketProps = config.WithCloudFront
```
normj: AWS AppSec doesn't really want us encouraging users to use the S3 website feature directly and instead do the CloudFront flow you have. For this PR we should make CloudFront the only option for best security practices.

src/Aspire.Hosting.AWS/Deployment/CDKPublishTargets/S3StaticWebsitePublishTarget.cs
```diff
@@ -0,0 +1,216 @@
...
+            ? new BucketProps
+            {
+                // Private bucket — CloudFront accesses via OAC
+                BlockPublicAccess = BlockPublicAccess.BLOCK_ALL,
```
normj: Can you follow the pattern used in the other publish targets and put all of the defaults in the `CDKDefaultsProvider` and then call ApplyX methods here. That gives us a mechanism of versioning the defaults we use.

src/Aspire.Hosting.AWS/Deployment/CDKPublishTargets/S3StaticWebsitePublishTarget.cs
```diff
@@ -0,0 +1,216 @@
...
+                    if (linkedAnnotation.Construct is ApplicationLoadBalancedFargateService albService)
+                        albDns = Token.AsString(albService.LoadBalancer.LoadBalancerDnsName);
+                    else
+                        throw new InvalidOperationException(
+                            $"Backend resource '{backendBehavior.BackendResource.Name}' (construct: {linkedAnnotation.Construct.GetType().Name}) " +
+                            $"is not supported as a CloudFront backend behavior origin. Only ECS Fargate with ALB is currently supported.");
+
+                    var behavior = new BehaviorOptions
+                    {
+                        Origin = new HttpOrigin(albDns, new HttpOriginProps
```
normj: Again all of these default values should from the CDKDefaultsProvider.

src/Aspire.Hosting.AWS/Deployment/CDKPublishTargets/S3StaticWebsitePublishTarget.cs
```diff
@@ -0,0 +1,216 @@
...
+                var additionalBehaviors = new Dictionary<string, IBehaviorOptions>();
+                foreach (var backendBehavior in config.BackendBehaviors)
+                {
+                    if (!backendBehavior.BackendResource.TryGetLastAnnotation<AWSLinkedObjectsAnnotation>(out var linkedAnnotation))
+                        throw new InvalidOperationException(
+                            $"Backend resource '{backendBehavior.BackendResource.Name}' has not been published yet. " +
+                            $"Ensure it is defined before the static website resource in your AppHost.");
+
+                    string albDns;
+                    if (linkedAnnotation.Construct is ApplicationLoadBalancedFargateService albService)
```
normj: Any reason the ECS Express is not supported? You can probably make this more generic by calling `linkAnnotation.PublishTarget.GetReferenceConnectionInfo(linkAnnotation)` and then looking for a `services__{linkedAnnotation.Resource.Name}__https__0` or `services__{linkedAnnotation.Resource.Name}__http__0` in the collection.

src/Aspire.Hosting.AWS/Deployment/CDKPublishTargets/S3StaticWebsitePublishTarget.cs
```diff
@@ -0,0 +1,216 @@
...
+        }
+
+        config.PropsBucketDeploymentCallback?.Invoke(context, deploymentProps);
+        new BucketDeployment(environment.CDKStack, $"Project-{resource.Name}-Deployment", deploymentProps);
+
+        ApplyAWSLinkedObjectsAnnotation(environment, resource, distribution ?? (Constructs.Construct)bucket, this);
+    }
+
+    public override ReferenceConnectionInfo GetReferenceConnectionInfo(AWSLinkedObjectsAnnotation linkedAnnotation)
+        => new();
```
normj: You should fill out this out like this one https://github.com/aws/integrations-on-dotnet-aspire-for-aws/blob/main/src/Aspire.Hosting.AWS/Deployment/CDKPublishTargets/ECSFargateExpressServicePublishTarget.cs#L96 but instead pull the endpoint from the distribution. That way other projects can have references on the Javascript project and find it via service discovery.

src/Aspire.Hosting.AWS/Deployment/PublishS3StaticWebsiteAnnotation.cs
```diff
@@ -0,0 +1,108 @@
...
+using Amazon.CDK.AWS.CloudFront;
+using Amazon.CDK.AWS.S3;
+using Amazon.CDK.AWS.S3.Deployment;
+using Aspire.Hosting.ApplicationModel;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Aspire.Hosting.AWS.Deployment;
+
+[Experimental(Constants.ASPIREAWSPUBLISHERS001)]
+internal class PublishS3StaticWebsiteAnnotation : IAWSPublishTargetAnnotation
```
normj: Per my suggestion above that this should sole CloudFront fronted focus the name should be updated to match that. Maybe `PublishS3WithCloudFrontAnnotation`.

src/Aspire.Hosting.AWS/Deployment/PublishS3StaticWebsiteAnnotation.cs
```diff
@@ -0,0 +1,108 @@
...
+    /// output. Defaults to <c>dist</c>. Angular apps typically use <c>dist/browser</c>.
+    /// </summary>
+    public string OutputPath { get; set; } = "dist";
+
+    /// <summary>
+    /// Backend behaviors that add a CloudFront origin for a specific path pattern, routing
+    /// matched requests to a backend resource (e.g. an ECS Fargate service with ALB) instead
+    /// of serving them from S3. Only used when <see cref="WithCloudFront"/> is <see langword="true"/>.
+    /// </summary>
+    public List<BackendBehavior> BackendBehaviors { get; } = [];
```
normj: Is there a reason we need this property vs rely on the `WithReference` calls done in the AppHost. That is how the existing connections are made between apps and don't want to invent a new mechanism.

normj (changes requested):
This is really cool to add Javascript support and connecting it to .NET backends. Can you update the Publishing playground app to also have a small Javascript app that uses this feature. Makes it easier to test the feel of this feature.

src/Aspire.Hosting.AWS/Deployment/PublishS3StaticWebsiteAnnotation.cs
halllo: It needs a way to provide the `pathPattern` used to route to the reference.
With the latest change I introduce `WithCloudFrontBackendBehavior("/todos/*", backend)`.
We could create an overload like `.WithReference(backend, "/todos/*")` but I think it is important to be transparent that it is only supported for the CloudFront integration.

src/Aspire.Hosting.AWS/Deployment/Services/DefaultStaticSiteBuilder.cs
```diff
@@ -0,0 +1,48 @@
...
+namespace Aspire.Hosting.AWS.Deployment.Services;
+
+internal class DefaultStaticSiteBuilder(IProcessCommandService processCommandService, ILogger<DefaultStaticSiteBuilder> logger) : IStaticSiteBuilder
+{
+    public Task BuildAsync(IResource resource, string workingDirectory, IDictionary<string, string> environmentVariables, CancellationToken cancellationToken)
+    {
+        cancellationToken.ThrowIfCancellationRequested();
+        var executableName = "npm";
+        var scriptCommand = "run";
+        if (resource.TryGetLastAnnotation<JavaScriptPackageManagerAnnotation>(out var pkgManager))
```
ndoherty48, line 17: What would be the approach, if looking to deploy a static side that is not a JavaScript app such as a Blazor WASM application for example? Would the end user just implement a `IStaticSiteBuilder` specific to that non JS env and add to DI to override this one?

Just curious about how this could work for a non JS static site?

PR #207, opened 2026-06-10
feat: add support for AWS.AgentCore.Hosting

philasmar:
## Summary

  Adds experimental support for integrating `AWS.AgentCore.Hosting` and `AWS.AgentCore.Testing` into the Aspire hosting experience. All emulators run as embedded in-process Kestrel servers — no Docker or separate processes required.

  ### New APIs (behind `ASPIREAWSAGENTCORE001` experimental flag)

  - **`AddAgentCoreRuntime<TProject>()`** — Registers an agent project with a dedicated runtime emulator and chat app. OS-assigned ports, automatic lifecycle management.
  - **`WithStreaming()`** — Configures the chat app to use SSE streaming mode.
  - **`WithInMemory()`** — Adds an embedded memory emulator for short-term memory.
  - **`WithReference(agent)`** — Injects `AWS_ENDPOINT_URL_BEDROCK_AGENTCORE` into a consumer project so the AWS SDK routes to the local emulator automatically. Falls through to native Aspire service discovery for non-AgentCore resources.

  ### Infrastructure

  - Multi-targets `Aspire.Hosting.AWS` to `net8.0;net10.0` — AgentCore extensions are `#if NET10_0_OR_GREATER` guarded.
  - Emulator error handling: try/catch with `TrySetException` propagation so downstream awaits fail fast instead of hanging.
  - Graceful shutdown via `ResourceStoppedEvent` subscription.

  ### Tests

  - 16 unit tests covering all public methods (happy path + error paths).
  - Integration tests that start the AppHost and verify agents respond to `/ping`.
  - Multi-targeted test projects (`net8.0` + `net10.0`), 195 existing tests unaffected.

  ### Apps

  - **`testapps/AgentCoreTestApps/`** — Minimal apps for automated test infrastructure (non-streaming agent, streaming agent, ChatUI, AppHost).
  - **`playground/AgentCore/`** — Developer demo showcasing all features: annotations-based agent, extension-method streaming agent, ChatUI with `WithReference`, `WithInMemory`, `IncludeEmulatorLogs`.

  ## Test plan

  - [x] All 16 AgentCore unit tests pass (`net10.0`)
  - [x] All 195 existing unit tests pass (`net8.0`, no regression)
  - [x] Integration tests build for both TFMs
  - [x] Playground AppHost launches, agents respond, chat app works end-to-end
  - [x] `WithReference` correctly falls through to native service discovery for non-AgentCore resources


README.md
```diff
@@ -170,6 +171,67 @@ For advanced deployment scenarios and comprehensive implementation details, see
...
 
+## Local Development with AgentCore (Experimental)
+
+> **Note**: This feature requires `net10.0` and is behind the `ASPIREAWSAGENTCORE001` experimental flag.
+
+The AgentCore integration provides a local development experience for agents built with the `AWS.AgentCore.Hosting` package. All emulators run as embedded in-process Kestrel servers — no Docker or separate processes required.
+
+### Getting Started
+
+Add the `AWS.AgentCore.Hosting` package to your agent project and use the annotations or extension methods to define your agent. Then in your Aspire AppHost:
```
GarrettBeatty: super nit: add `dotnet add package` here

src/Aspire.Hosting.AWS/AgentCore/AgentCoreRuntimeAnnotation.cs
```diff
@@ -0,0 +1,47 @@
...
+            try
+            {
+                await server.StopAsync();
+            }
+            catch
+            {
+                // Best-effort shutdown
+            }
+        }
+    }
```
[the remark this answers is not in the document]
GarrettBeatty: check this

src/Aspire.Hosting.AWS/AgentCore/AgentCoreResourceBuilderExtensions.cs
```diff
@@ -0,0 +1,280 @@
...
+    /// emulator endpoint as the <c>AWS_ENDPOINT_URL_BEDROCK_AGENTCORE</c> environment variable.
+    /// </summary>
+    /// <typeparam name="TProject">The agent project type.</typeparam>
+    /// <param name="builder">The distributed application builder.</param>
+    /// <param name="name">Optional resource name. Defaults to the project type name.</param>
+    /// <param name="options">Optional configuration for the local development experience.</param>
+    /// <returns>The agent's resource builder for further configuration.</returns>
+    public static IResourceBuilder<ProjectResource> AddAgentCoreRuntime<TProject>(
+        this IDistributedApplicationBuilder builder,
+        string? name = null,
```
normj: The convention is to have `name` required especially for the `AddProject` which I view this method should be similar to. We shouldn't create a new convention and generate defaults. Also `name` parameters used as resource names have the `[ResourceName]` parameter attribute applied to it.

src/Aspire.Hosting.AWS/AgentCore/AgentCoreResourceBuilderExtensions.cs
```diff
@@ -0,0 +1,280 @@
...
+    /// Returns the agent's <see cref="IResourceBuilder{ProjectResource}"/>.
+    /// Use <c>.WithReference(agent)</c> from another project to inject the runtime
+    /// emulator endpoint as the <c>AWS_ENDPOINT_URL_BEDROCK_AGENTCORE</c> environment variable.
+    /// </summary>
+    /// <typeparam name="TProject">The agent project type.</typeparam>
+    /// <param name="builder">The distributed application builder.</param>
+    /// <param name="name">Optional resource name. Defaults to the project type name.</param>
+    /// <param name="options">Optional configuration for the local development experience.</param>
+    /// <returns>The agent's resource builder for further configuration.</returns>
+    public static IResourceBuilder<ProjectResource> AddAgentCoreRuntime<TProject>(
```
normj, line 33: Look here how we did Lambda to return a subclass of `ProjectResource` like `AgentCoreProjectResource`. Return the subclass will help use have more targeted extension methods.

src/Aspire.Hosting.AWS/AgentCore/AgentCoreResourceBuilderExtensions.cs
```diff
@@ -0,0 +1,280 @@
...
+        // This guarantees ports are known before env var callbacks fire.
+        builder.Eventing.Subscribe<BeforeResourceStartedEvent>(
+            agentApp.Resource,
+            async (@event, ct) =>
+            {
+                try
+                {
+                    var loggerService = @event.Services.GetRequiredService<ResourceLoggerService>();
+
+                    var agentEndpointUrl = "http://localhost:8080";
```
normj: I understand in AgentCore we have to use port 8080 but locally we can't hard code the port to 8080. It might be being used by something else also it restricts the ability to have multiple agents running in Aspire.

src/Aspire.Hosting.AWS/AgentCore/AgentCoreResourceBuilderExtensions.cs
```diff
@@ -0,0 +1,280 @@
...
+
+            await annotation.EmulatorStarted.Task;
+            context.EnvironmentVariables["AWS_AGENTCORE_SERVICE_ENDPOINT"] =
+                $"http://localhost:{annotation.MemoryPort}";
+        });
+
+        return agentApp;
+    }
+
+    private static int GetBoundPort(Microsoft.AspNetCore.Builder.WebApplication app)
```
normj, line 266: Looks like you are recreating logic for assigning ports instead of relying on Aspire's built in mechanism. Here is the code how we allocate a port for the Lambda emulator. https://github.com/aws/integrations-on-dotnet-aspire-for-aws/blob/main/src/Aspire.Hosting.AWS/Lambda/LambdaExtensions.cs#L159

We also pretty quickly with Lambda got customer ask for the ability in the AppHost to configure the emulator port.

philasmar: The Lambda emulator pattern (EndpointAnnotation → DCP allocates port → process binds to it) doesn't directly apply here because the AgentCore emulators are in-process Kestrel servers running inside the AppHost, not separate executable processes managed by DCP.

With Lambda, LambdaEmulatorResource extends ExecutableResource — DCP launches it as a separate process and can allocate/manage its ports. Our runtime emulator, chat app, and memory emulator are embedded WebApplication instances started in the BeforeResourceStartedEvent handler. If we add EndpointAnnotations to the agent resource for the emulators, DCP interprets that as "this process should also bind to these ports," which causes the agent to conflict with the emulators on the same ports.

To address the configurable port ask (which Lambda customers also needed), I've added RuntimeEmulatorPort and ChatAppPort to AgentCoreLocalOptions:

```
builder.AddAgentCoreRuntime<Projects.MyAgent>("my-agent", new AgentCoreLocalOptions
{
    RuntimeEmulatorPort = 9000,
    MemoryEmulatorPort = 9001,
    ChatAppPort = 9002
});
```

When null (default), ports are OS-assigned. This gives the same user-facing control as Lambda's HttpPort/HttpsPort options.

If we want to fully align with the Lambda model in the future, we could extract the emulators into separate ExecutableResources managed by DCP — but that would require packaging them as standalone executables rather than library methods, which is a larger architectural change in AWS.AgentCore.Testing.

src/Aspire.Hosting.AWS/AgentCore/AgentCoreResourceBuilderExtensions.cs
philasmar: Done

src/Aspire.Hosting.AWS/AgentCore/AgentCoreResourceBuilderExtensions.cs
philasmar: I had originally done this but ran into issues that come with us not using AddProject in the backend. I went back to that approach.

src/Aspire.Hosting.AWS/AgentCore/AgentCoreResourceBuilderExtensions.cs
philasmar: Fixed

src/Aspire.Hosting.AWS/AgentCore/AgentCoreRuntimeAnnotation.cs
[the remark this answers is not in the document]
philasmar: Done

README.md
philasmar: Done

GarrettBeatty (approved):

src/Aspire.Hosting.AWS/AgentCore/AgentCoreResourceBuilderExtensions.cs
```diff
@@ -0,0 +1,284 @@
...
+            context.Urls.RemoveAll(u =>
+                u.Endpoint is not null &&
+                u.DisplayText != "Chat" &&
+                u.DisplayText != "Runtime Emulator" &&
+                u.DisplayText != "Agent Instance");
+        });
+
+        var annotation = new AgentCoreRuntimeAnnotation
+        {
+            IncludeEmulatorLogs = options.IncludeEmulatorLogs
```
normj: The annotation has all the properties for the ports, shouldn't we copy them from the options class to the annotation?

src/Aspire.Hosting.AWS/AgentCore/AgentCoreResourceBuilderExtensions.cs
```diff
@@ -0,0 +1,284 @@
...
+    {
+        ArgumentNullException.ThrowIfNull(builder);
+        ArgumentNullException.ThrowIfNull(name);
+
+        options ??= new AgentCoreLocalOptions();
+
+        var projectName = name;
+
+        var project = new AgentCoreRuntimeResource(projectName);
+        var agentApp = builder.AddResource(project)
```
normj: As much as I prefer having our own subclass I think you are right by not using the `AddProject` we are setting ourselves up for issues skipping the extra logic `AddProject`. We should revert back to using `AddProject` and then the return type is `ProjectResource`.

playground/AgentCore/AgentCore.AppHost/AgentCore.AppHost.csproj
```diff
@@ -0,0 +1,25 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
normj, line 1: The apphost project needs a launchsettings.json. Otherwise launching the apphost fails.

tests/Aspire.Hosting.AWS.Integ.Tests/Aspire.Hosting.AWS.Integ.Tests.csproj
```diff
@@ -1,19 +1,28 @@
...
+		<ProjectReference Include="..\..\playground\Lambda\Lambda.AppHost\Lambda.AppHost.csproj" Condition="'$(TargetFramework)' == '$(DefaultTargetFramework)'" />
+		<ProjectReference Include="..\..\playground\CloudFormationProvisioning\AWS.AppHost\AWS.AppHost.csproj" Condition="'$(TargetFramework)' == '$(DefaultTargetFramework)'" />
 		<ProjectReference Include="..\..\src\Aspire.Hosting.AWS\Aspire.Hosting.AWS.csproj" />
-		<ProjectReference Include="..\..\testapps\DeploymentTestApps\DeploymentTestApp.AppHost\DeploymentTestApp.AppHost.csproj" />
+		<ProjectReference Include="..\..\testapps\DeploymentTestApps\DeploymentTestApp.AppHost\DeploymentTestApp.AppHost.csproj" Condition="'$(TargetFramework)' == '$(DefaultTargetFramework)'" />
+		<ProjectReference Include="..\..\testapps\AgentCoreTestApps\AgentCoreTestApp.AppHost\AgentCoreTestApp.AppHost.csproj" Condition="'$(TargetFramework)' == '$(AgentCoreTargetFramework)'" />
+	</ItemGroup>
+
+	<!-- Exclude non-AgentCore tests from net10.0 build (CDK/JSII not compatible with net10.0) -->
+	<ItemGroup Condition="'$(TargetFramework)' == '$(AgentCoreTargetFramework)'">
```
normj: We need to figure out what is going wrong with the tests because we know CDK/JSII is compatible with .NET 10. At a minimum .NET 10 should be are main target but it would be good to understand what is going wrong with the tests running both targets.

Directory.Build.props
```diff
@@ -2,6 +2,7 @@
   
   <PropertyGroup>
     <DefaultTargetFramework>net8.0</DefaultTargetFramework>
+    <AgentCoreTargetFramework>net10.0</AgentCoreTargetFramework>
```
normj: Once you figure out what is going wrong with the tests it seems like we shouldn't need a separate `AgentCoreTargetFramework` property and could just have `DefaultTargetFrameworks` equal `net8.0;net10.0`.

Directory.Packages.props
```diff
@@ -20,18 +20,21 @@
...
     <PackageVersion Include="AWSSDK.DynamoDBv2" Version="4.0.18.1" />
     <PackageVersion Include="AWSSDK.Lambda" Version="4.0.15.2" />
-    <PackageVersion Include="AWSSDK.S3" Version="4.0.21.2" />
     <PackageVersion Include="AWSSDK.SecurityToken" Version="4.0.6.3" />
     <PackageVersion Include="AWSSDK.SimpleNotificationService" Version="4.0.2.30" />
     <PackageVersion Include="AWSSDK.SQS" Version="4.0.2.28" />
     <PackageVersion Include="AWSSDK.SSO" Version="4.0.2.27" />
     <PackageVersion Include="AWSSDK.SSOOIDC" Version="4.0.3.28" />
-    <PackageVersion Include="AWSSDK.Extensions.NETCore.Setup" Version="4.0.3.37" />
+    <PackageVersion Include="AWSSDK.Extensions.NETCore.Setup" Version="4.0.3.40" />
```
normj, line 38: Unrelated to your change but since you are already making modifications here can you add `AWSSDK.Signin` so that we support credentials that were vended via `aws login`?

README.md
```diff
@@ -170,6 +171,73 @@ For advanced deployment scenarios and comprehensive implementation details, see
 
 The design document provides detailed examples and implementation guidance for these advanced use cases, enabling you to customize deployments for production workloads.
 
+## Local Development with AgentCore (Experimental)
```
normj, line 174: We actually have 2 README files. The root one should be more of quick summary and then there is the `src/Aspire.Hosting.AWS/README.md` which is packaged in the package and is more indepth. Some day we will probably have more then one package in the repo so we don't want to use the root readme file for everything.

tests/Aspire.Hosting.AWS.Integ.Tests/Aspire.Hosting.AWS.Integ.Tests.csproj
philasmar: When multi-targeting the integration test project to both net8.0 and net10.0, the CDK deployment tests that use Vpc.FromLookup failed on net10.0 with unresolved placeholder values (dummy1a, vpc-12345) in CloudFormation templates. I initially suspected a JSII/.NET 10 incompatibility, but the root cause was entirely different.

The DeploymentTestApp.AppHost targeted net8.0 while the test runner ran under net10.0.

When DistributedApplicationTestingBuilder.CreateAsync<Projects.DeploymentTestApp_AppHost>() runs, it loads the AppHost code in-process. The CDK context generation step (GetCDKContext) forks a child process:

```
  var fullPath = Assembly.GetEntryAssembly()!.Location;
  var appHostAssembly = Config.OverrideAppHostAssemblyName ?? Path.GetFileName(fullPath);
  string workingDirectory = Directory.GetParent(fullPath)!.FullName;
  // ...
  processCommandService.RunCDKProcess(..., $"--app 'dotnet exec {appHostAssembly}' synth ...", workingDirectory, ...);
```

  Under the net10.0 test host, Assembly.GetEntryAssembly()!.Location points to the test host DLL in the net10.0/ output directory. The working directory becomes bin/Debug/net10.0/. But DeploymentTestApp.AppHost.dll in that directory was compiled for net8.0 (copied via project reference). When dotnet exec runs it, the net8.0 runtimeconfig.json rolls forward to the net10.0 runtime, but the CDK context generation subprocess encounters environment differences that cause it to silently fail (return 
  null), so CDK synthesizes templates with placeholder values.

I switched everything to .NET10 except for Aspire.Hosting.AWS which is multi targeted.

Directory.Build.props
philasmar: Since I switched everything to .NET10 except for Aspire.Hosting.AWS, I made DefaultTargetFramework = net10 and removed AgentCoreTargetFramework

src/Aspire.Hosting.AWS/AgentCore/AgentCoreResourceBuilderExtensions.cs
philasmar: Fixed

src/Aspire.Hosting.AWS/AgentCore/AgentCoreResourceBuilderExtensions.cs
philasmar: Fixed

playground/AgentCore/AgentCore.AppHost/AgentCore.AppHost.csproj
philasmar: Fixed

Directory.Packages.props
philasmar: Fixed

README.md
philasmar: Fixed

normj (approved):

PR #214, opened 2026-07-01
Add AgentCore memory support when deploying

normj:
## Summary

Adds provisioning of AWS Bedrock **AgentCore Memory** during deployment. Previously `WithInMemory()` only stood up a local memory emulator; none of that translated to deployed infrastructure. Now, when memory is requested on an agent, deployment creates an `AWS::BedrockAgentCore::Memory` resource, points the agent at it, and grants the agent's IAM role access to it.

## ⚠️ Breaking change — `WithInMemory` renamed to `WithAgentCoreMemory`

The experimental extension method **`WithInMemory()` has been renamed to `WithAgentCoreMemory()`**. Because it is exposed on the generic `IResourceBuilder<ProjectResource>` (shared with non-agent projects), the new name makes it clear the method configures AgentCore memory. Behavior for local development is unchanged. This is an experimental (`ASPIREAWSAGENTCORE001`) API, so it was renamed outright rather than kept as an alias — update any `.WithInMemory()` calls to `.WithAgentCoreMemory()`.

## What's included

**Deployment: create the memory resource**
- When `WithAgentCoreMemory()` was called (`AgentCoreRuntimeAnnotation.HasMemory`), `AgentCoreRuntimePublishTarget` now creates a `CfnMemory` (`AWS::BedrockAgentCore::Memory`) construct following the existing publish-target conventions.
- A short-term (event) memory is created — matching what the local emulator provides. It defaults an `EventExpiryDuration` (90 days) and derives a sanitized memory name from the stack + resource name. Long-term memory strategies / a memory execution role are intentionally left to the user via the new props callback.

**Agent wiring**
- The agent's `AWS_AGENTCORE_MEMORY_ID` environment variable is pointed at the created memory's id (`Fn::GetAtt … MemoryId`), replacing the local `localdev-memory` placeholder during publish.
- A dependency is added so the memory is provisioned before the runtime.

**IAM**
- When the agent's runtime role was created by the integration (not user-supplied), it is granted the AgentCore memory **data-plane** permissions (`CreateEvent`, `GetEvent`, `ListEvents`, `DeleteEvent`, `ListActors`, `ListSessions`, `RetrieveMemoryRecords`, `GetMemoryRecord`, `ListMemoryRecords`), scoped to the created memory's ARN and its child resources. User-supplied roles are never mutated.

**Customization (follows the publishing-target callback pattern)**
- `PublishAgentCoreRuntimeConfig.PropsCfnMemoryCallback` — customize the memory props before construction.
- `PublishAgentCoreRuntimeConfig.ConstructCfnMemoryCallback` — customize the constructed memory.
- `PublishAgentCoreRuntimeConfig.CreateMemory` (`bool?`) — opt-out/override. Defaults to following `WithAgentCoreMemory()`; set `false` to use the emulator locally while skipping memory provisioning on deploy (resolved as `CreateMemory ?? HasMemory`, so it works regardless of call order).

## Tests

- New deployment scenarios `PublishAgentCoreRuntimeWithMemory` and `PublishAgentCoreRuntimeWithMemoryDisabled`.
- New CloudFormation assertion tests:
  - `TestPublishAgentCoreRuntimeWithMemory` — asserts a single `AWS::BedrockAgentCore::Memory` is emitted, `AWS_AGENTCORE_MEMORY_ID` resolves via `Fn::GetAtt` to the memory (not `localdev-memory`), and the runtime role has the memory data-plane policy scoped to the memory ARN.
  - `TestPublishAgentCoreRuntimeWithMemoryDisabled` — asserts no memory resource is emitted when `CreateMemory = false`.
- Updated existing AgentCore unit tests for the rename.

## Playground

- `WithInMemory()` → `WithAgentCoreMemory()` in the AgentCore and AgentCoreTestApp app hosts.
- The Publishing playground's HoroscopeAgent now uses AgentCore memory (`WithAgentCoreMemory()`) in place of the Redis cache.

## Notes

- All new CDK/deployment code is under `#if NET10_0_OR_GREATER`, matching the rest of the AgentCore publish target.
- `CfnMemoryProps.EventExpiryDuration` is a non-nullable `double`, so `0` is treated as "unset" when applying the default.

### DIFF
--- a/.autover/changes/ed8515cc-43f4-43b6-85a7-3b77f26a8ed3.json
+++ b/.autover/changes/ed8515cc-43f4-43b6-85a7-3b77f26a8ed3.json
@@ -1,0 +1,12 @@
+{
+  "Projects": [
+    {
+      "Name": "Aspire.Hosting.AWS",
+      "Type": "Patch",
+      "ChangelogMessages": [
+        "When deploying to AgentCore support provisioning AgentCore memory",
+		"[Breaking Change] Rename the preview WithInMemory method to WithAgentCoreMemory"
+      ]
+    }
+  ]
+}
--- a/.autover/changes/ed8515cc-43f4-43b6-85a7-3b77f26a8ed3.json
+++ b/.autover/changes/ed8515cc-43f4-43b6-85a7-3b77f26a8ed3.json
@@ -2,7 +2,7 @@
   "Projects": [
     {
       "Name": "Aspire.Hosting.AWS",
-      "Type": "Patch",
+      "Type": "Minor",
       "ChangelogMessages": [
         "When deploying to AgentCore support provisioning AgentCore memory",
 		"[Breaking Change] Rename the preview WithInMemory method to WithAgentCoreMemory"
--- a/src/Aspire.Hosting.AWS/Deployment/CDKDefaults/CDKDefaultsProvider.AgentCoreRuntime.cs
+++ b/src/Aspire.Hosting.AWS/Deployment/CDKDefaults/CDKDefaultsProvider.AgentCoreRuntime.cs
@@ -19,6 +19,7 @@ public partial class CDKDefaultsProvider
     public virtual string AgentCoreRuntimeNetworkMode => "PUBLIC";
 
     private IRole? _defaultAgentCoreRuntimeRole;
+    private bool _defaultAgentCoreRuntimeRoleCreatedByIntegration;
 
     /// <summary>
     /// Gets the default IAM role for AgentCore runtimes. The role is shared across all AgentCore runtimes in the stack.
@@ -32,10 +33,12 @@ public partial class CDKDefaultsProvider
             if (definedDefault != null)
             {
                 _defaultAgentCoreRuntimeRole = definedDefault;
+                _defaultAgentCoreRuntimeRoleCreatedByIntegration = false;
             }
             else
             {
                 _defaultAgentCoreRuntimeRole = CreateDefaultAgentCoreRuntimeRole();
+                _defaultAgentCoreRuntimeRoleCreatedByIntegration = true;
             }
         }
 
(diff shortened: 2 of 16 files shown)


src/Aspire.Hosting.AWS/Deployment/CDKDefaults/CDKDefaultsProvider.AgentCoreRuntime.cs
```diff
@@ -114,6 +117,71 @@ protected internal virtual void ApplyAgentCoreRuntimeDefaults(CfnRuntimeProps pr
...
+    /// </para>
+    /// </summary>
+    /// <param name="memoryArn">The ARN of the AgentCore memory to grant access to.</param>
+    internal void GrantAgentCoreRuntimeMemoryAccess(string memoryArn)
+    {
+        var role = GetDefaultAgentCoreRuntimeRole();
+        if (!_defaultAgentCoreRuntimeRoleCreatedByIntegration)
+        {
+            return;
+        }
```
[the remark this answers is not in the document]
normj: This follows the pattern we have already been using in Aspire. I did some tweaks so that if multiple Memory instances are created that we don't duplicate the IAM actions in the role's policy.

GarrettBeatty (approved):
shouldnt there be a change file. can you add one with breaking change not in changelog

.autover/changes/ed8515cc-43f4-43b6-85a7-3b77f26a8ed3.json
```diff
@@ -0,0 +1,12 @@
+{
+  "Projects": [
+    {
+      "Name": "Aspire.Hosting.AWS",
+      "Type": "Patch",
```
philasmar: We should make this a Minor given the breaking change

normj:
> shouldnt there be a change file. can you add one with breaking change not in changelog

Just added a change file

.autover/changes/ed8515cc-43f4-43b6-85a7-3b77f26a8ed3.json
normj: Done

philasmar (approved):

PR #219, opened 2026-07-24
[Breaking Change] Switch ECS Express to use TaskDefinition Instead of PrimaryContainer

normj:
## Summary

Reworks the **ECS Fargate Express Service** deployment target so it runs the project on a real ECS **`FargateTaskDefinition`** and points the `CfnExpressGatewayService` at it via the new `TaskDefinitionArn` property — instead of describing the container inline through the `PrimaryContainer` property.

This brings the Express target in line with the other ECS Fargate publish targets (which already use a `FargateTaskDefinition`) and, critically, unblocks future work: **adding an OpenTelemetry side-car container**, which must be added to a task definition alongside the application container. The inline `PrimaryContainer` model has no place to attach a second container.

## ⚠️ Breaking Change

This is a **known, intentional breaking change** to the generated CloudFormation for web projects deployed via `PublishAsECSFargateExpressService` (the default for web `ProjectResource`s):

- Aspire deployment is currently **marked as preview** requiring users to opt-in to allow for breaking changes like this.
- The Express service no longer carries an inline `PrimaryContainer`. Instead a new `AWS::ECS::TaskDefinition` resource is emitted per Express service and referenced by the service's `TaskDefinitionArn`.
- Container-hosting settings move from the Express service props onto the task definition / its container: **image, environment variables, container port, CPU, memory, execution role, and task role**.
- Because a task definition is now created (with an awslogs log driver), each Express service also emits a **log group** and the execution role gains a **default policy** (ECR pull + log writes).

Redeploying an existing stack will replace the Express service's container wiring. This is required for the OpenTelemetry side-car work noted above and is captured in the changelog as a breaking change. 

## Changes

- **`ECSFargateExpressServicePublishTarget`** — `GenerateConstructAsync` now creates a `FargateTaskDefinition`, adds the application container via `ContainerImage.FromTarball(...)`, and sets `CfnExpressGatewayServiceProps.TaskDefinitionArn = taskDef.TaskDefinitionArn`. The reference/connection-points logic (env vars, security group, task role) now targets the container definition. Replaced `CfnExpressGatewayServicePropsConnectionPoints` with `ExpressServiceContainerConnectionPoints` (wraps `ContainerDefinitionProps`).
- **ECS Express requirements** — the application container is named **`Main`** and its port mapping is a **named TCP port** (`ContainerPort` 8080, `Name` `http`, `Protocol` TCP). ECS Fargate Express rejects the task definition without both.
- **`CDKDefaultsProvider.ECSFargateExpressService`** — split the single `ApplyCfnExpressGatewayServiceDefaults` into three: task-definition defaults (CPU / memory / execution role), container defaults (awslogs log driver + named TCP port mapping), and slimmed service defaults (infrastructure role + network configuration). Added the `ECSFargateExpressContainerPortName` virtual (default `"http"`); reused the existing `ECSFargateExpressCpu` / `ECSFargateExpressMiB` / `ECSFargateExpressContainerPort` virtuals and the shared `CreateECSFargateServiceLogDriver`.
- **`PublishECSFargateExpressServiceConfig`** — added `FargateTaskDefinition` and `ContainerDefinition` props/construct callbacks (full parity with `PublishAsECSFargateService`), keeping the existing Express-service callbacks. Additive and non-breaking to the config surface.
- **`Amazon.CDK.Lib`** bumped `2.236.0` → `2.262.0` (the `TaskDefinitionArn` property on `CfnExpressGatewayServiceProps` was introduced in 2.261.0).
- **Tests** — updated `ApplyDefaultsTests` (split into the three new methods; asserts named TCP port) and `PublishScenarioTests` (relocated container/CPU/memory/execution-role/task-role assertions to the new task-definition resource; container name lookups use `Main`; task-def / log-group / policy counts updated). The `PublishWebApp2ReferenceOnWebApp1` scenario now sets memory via the new `PropsFargateTaskDefinitionCallback`.
- **`docs/deployment-design.md`** — updated the ECS task-role and defaults sections, and added an "ECS Fargate Express Task Definition Requirements" section documenting the `Main` container name and named-TCP-port requirements.

## Tooling note

CDK lib 2.262.0 emits cloud-assembly schema 54, which requires a CDK **CLI** newer than 2.1106.0 (max schema 52). The CLI must be at 2.1108.0+ (2.1133.0 verified) for the publish/deploy context-generation step; otherwise VPC/AZ lookups fall back to placeholders.

## Testing

- Unit tests: `ApplyDefaultsTests` pass (20/20).
- Integration tests (CDK synthesis): all Express scenarios pass — WebApp2Reference, Service1Reference, WithEnvironment(+Reference), AgentCoreRuntimeWithReference, Serverless/Provisioned Redis & Valkey, WebApp1UsingDefaultVpc.
- Verified end-to-end deploy: the synthesized task definition has the `Main` container with a named TCP port mapping and the Express service references it via `TaskDefinitionArn`.

### DIFF
--- a/src/Aspire.Hosting.AWS/Deployment/CDKDefaults/CDKDefaultsProvider.ECSFargateExpressService.cs
+++ b/src/Aspire.Hosting.AWS/Deployment/CDKDefaults/CDKDefaultsProvider.ECSFargateExpressService.cs
@@ -31,35 +31,65 @@ public partial class CDKDefaultsProvider
     public virtual double? ECSFargateExpressContainerPort => 8080;
 
     /// <summary>
+    /// Gets the name given to the default container port mapping. ECS Fargate Express requires the primary
+    /// ("Main") container to have a named TCP port mapping.
+    /// </summary>
+    /// <remarks>
+    /// Default is "http".
+    /// </remarks>
+    public virtual string ECSFargateExpressContainerPortName => "http";
+
+    /// <summary>
+    /// Applies default values to the Fargate task definition used by an ECS Express Gateway service, if they
+    /// are not already set. The Express service points at this task definition via its TaskDefinitionArn
+    /// property, so container-hosting settings such as CPU, memory, and the execution role live here.
+    /// </summary>
+    /// <param name="props">The Fargate task definition properties to which default values will be applied.</param>
+    protected internal virtual void ApplyCfnExpressGatewayServiceTaskDefinitionDefaults(FargateTaskDefinitionProps props)
+    {
+        if (props.Cpu == null)
+            props.Cpu = ECSFargateExpressCpu;
+        if (props.MemoryLimitMiB == null)
+            props.MemoryLimitMiB = ECSFargateExpressMiB;
+        if (props.ExecutionRole == null)
+            props.ExecutionRole = GetDefaultECSExpressExecutionRole();
+    }
+
+    /// <summary>
+    /// Applies default values to the container definition run by an ECS Express Gateway service's task
+    /// definition, if they are not already set. This adds awslogs logging and the default container port.
+    /// </summary>
+    /// <param name="projectName">The name of the project, used to create the log stream prefix.</param>
+    /// <param name="props">The container definition properties to which default values will be applied.</param>
+    protected internal virtual void ApplyCfnExpressGatewayServiceContainerDefinitionDefaults(string projectName, ContainerDefinitionProps props)
+    {
+        if (props.Logging == null)
+            props.Logging = CreateECSFargateServiceLogDriver(projectName);
+
+        if (props.PortMappings == null || props.PortMappings.Length == 0)
+        {
+            // ECS Fargate Express requires the Main container to expose a named TCP port mapping.
+            props.PortMappings = new[]
+            {
+                new PortMapping
+                {
+                    ContainerPort = ECSFargateExpressContainerPort ?? 8080,
+                    Name = ECSFargateExpressContainerPortName,
+                    Protocol = Protocol.TCP
+                }
+            };
+        }
+    }
+
+    /// <summary>
     /// Applies default values to the specified properties for configuring an AWS CloudFormation Express Gateway
-    /// service, if they are not already set.
+    /// service, if they are not already set. Container-hosting settings (CPU, memory, image, execution role,
+    /// container port) live on the task definition; this method only sets the service-level defaults.
     /// </summary>
     /// <param name="props">The properties object to which default values will be applied. Properties that are null or empty will be set to
     /// recommended defaults for an Express Gateway service.</param>
-    /// <exception cref="InvalidDataException">Thrown if the <paramref name="props"/>.PrimaryContainer property is not set or is not of type <see
-    /// cref="CfnExpressGatewayService.ExpressGatewayContainerProperty"/>.</exception>
     protected internal virtual void ApplyCfnExpressGatewayServiceDefaults(CfnExpressGatewayServiceProps props)
     {
-        if (props.Cluster == null)
-            props.Cluster = GetDefaultECSCluster().ClusterName;
-        if (string.IsNullOrEmpty(props.Cpu))
-            props.Cpu = ECSFargateExpressCpu.ToString();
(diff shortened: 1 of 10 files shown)


src/Aspire.Hosting.AWS/Deployment/CDKDefaults/CDKDefaultsProvider.ECSFargateExpressService.cs
```diff
@@ -31,35 +31,65 @@ public partial class CDKDefaultsProvider
...
+            props.PortMappings = new[]
+            {
+                new PortMapping
+                {
+                    ContainerPort = ECSFargateExpressContainerPort ?? 8080,
+                    Name = ECSFargateExpressContainerPortName,
+                    Protocol = Protocol.TCP
+                }
+            };
         }
```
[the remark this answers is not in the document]
normj: Good catch — addressed in the latest commit. Since ECS Express requires the `Main` container to have a named TCP port mapping, I added synth-time validation (`ValidateExpressPortMappings`) that runs after the user's `PropsContainerDefinitionCallback`, mirroring the existing `ContainerName == "Main"` check:

- The **default path is unchanged** — when the user supplies no mappings, `ApplyCfnExpressGatewayServiceContainerDefinitionDefaults` still adds a compliant named TCP mapping.
- When the user **does** supply mappings, we validate that **at least one** is a named TCP port with a container port, and throw `InvalidOperationException` otherwise — failing at synth time instead of at deploy.

A couple of deliberate choices:
- The requirement is "at least one named TCP port", **not** position-dependent — a container can legitimately expose multiple ports, so the named TCP one need not be first in the list.
- A `null` protocol is treated as TCP, since ECS/CDK default an omitted protocol to TCP; rejecting it would fail configs that ECS actually accepts.

Added unit tests for `ValidateExpressPortMappings` covering: null/empty, named-TCP-not-first, null-protocol-as-TCP, unnamed port (throws), and non-TCP protocol (throws). The docs section on Express task-definition requirements was updated to match.

GarrettBeatty (approved):

src/Aspire.Hosting.AWS/Deployment/CDKDefaults/CDKDefaultsProvider.ECSFargateExpressService.cs
```diff
@@ -31,35 +31,65 @@ public partial class CDKDefaultsProvider
...
+    /// ("Main") container to have a named TCP port mapping.
     /// </summary>
-    /// <param name="props">The properties object to which default values will be applied. Properties that are null or empty will be set to
-    /// recommended defaults for an Express Gateway service.</param>
-    /// <exception cref="InvalidDataException">Thrown if the <paramref name="props"/>.PrimaryContainer property is not set or is not of type <see
-    /// cref="CfnExpressGatewayService.ExpressGatewayContainerProperty"/>.</exception>
-    protected internal virtual void ApplyCfnExpressGatewayServiceDefaults(CfnExpressGatewayServiceProps props)
-    {
-        if (props.Cluster == null)
-            props.Cluster = GetDefaultECSCluster().ClusterName;
```
dscpinheiro, line 44: One question: Is dropping this default cluster assignment intentional? It exists on the other ECS defaults class: https://github.com/aws/integrations-on-dotnet-aspire-for-aws/blob/e7cfed4c507aec8287686735e7fe17b28fc2b9ab/src/Aspire.Hosting.AWS/Deployment/CDKDefaults/CDKDefaultsProvider.ECSFargateServiceWithALB.cs#L82-L83 

And on the ~~L2 construct~~ CFN resource: https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-resource-ecs-expressgatewayservice.html

dscpinheiro (approved):

src/Aspire.Hosting.AWS/Deployment/CDKDefaults/CDKDefaultsProvider.ECSFargateExpressService.cs
normj: Great catch! Update the PR and made sure the test coverage was added back.

dscpinheiro (approved):

PR #221, opened 2026-07-27
Add support for Aspire.Hosting.AWS features for Node.js based Aspire AppHost

normj:
## Description

Adds support for using `Aspire.Hosting.AWS` features from a **Node.js (TypeScript) Aspire AppHost**. .NET Aspire lets the AppHost be authored in TypeScript instead of C#; this PR annotates the AWS integration's public APIs with the Aspire Type System (ATS) attributes so the Aspire CLI can generate a TypeScript SDK for them. Only the AppHost authoring language changes — the Lambda functions and AWS resources remain .NET based.

The following features are now usable from a Node.js AppHost:

- **AWS SDK configuration** — `addAWSSDKConfig` + `withProfile` / `withRegion` / `withSdkValidation`
- **DynamoDB Local** — `addAWSDynamoDBLocal`
- **Lambda functions** — `addAWSLambdaFunction` for both the executable and class library programming models
- **Lambda service emulator** — `addAWSLambdaServiceEmulator`
- **API Gateway emulator** — `addAWSAPIGatewayEmulator` fronting Lambda functions
- **Event sources** — `withSQSEventSource` and `withDynamoDBStreamsEventSource` (string overloads)

## How it works

The mechanism is the Aspire Type System (ATS). Public APIs are marked with `[AspireExport]` (methods become capabilities; resource types become opaque handles) and `[AspireDto]` (option POCOs become value objects). At `aspire run`, the CLI restores the integration NuGet package, scans the assembly for these attributes, generates `aspire.mts`, and proxies each TypeScript call back to the real .NET code over JSON-RPC.

## Key design decisions

- **`WithReference` renamed in TypeScript.** Core Aspire owns the `withReference` name in polyglot AppHosts, and the AWS package has multiple `WithReference` overloads that would collide. Each is exported under an explicit, stable name via `[AspireExport("...")]` so it never auto-suffixes (which would silently break callers):
  - `WithReference(IAWSSDKConfig)` → `withAWSSDKConfigReference`
  - `WithReference(DynamoDBLocalResource)` → `withDynamoDBLocalReference`
  - API Gateway `WithReference(lambda, Method, path)` → `withAPIGatewayLambdaReference`
- **AWS SDK `ConstantClass` types have no clean TS representation.** `WithRegion` gets a string overload (`"us-west-2"`); Lambda log options are exposed via an `internal` TS-only DTO (`LambdaFunctionPolyglotOptions`) with string properties, so .NET AppHosts keep using `LambdaFunctionOptions` unchanged.
- **Generic `AddAWSLambdaFunction<TProject>` can't be called from TS.** Mirroring core Aspire's `AddProjectForPolyglot`, an `internal` path-based overload is exported that takes a project path string, resolves it (relative → absolute, directory → `.csproj`), and builds the project during `dotnet run` (`SuppressBuild = false`).
- **Event sources** export only the string (queue URL / table name) overloads; the CDK-construct and CloudFormation overloads remain C#-only.

Full design writeup: `docs/polyglot-design.md`.

## Changes

- **ATS annotations** across the SDK config, DynamoDB Local, and Lambda/API Gateway/event-source APIs (`src/Aspire.Hosting.AWS/**`).
- **New polyglot Lambda entry point** and helpers in `LambdaExtensions.cs` (+ `LambdaProjectMetadata.SuppressBuild` made configurable in `LambdaBeforeStartEventHandler.cs`).
- **`docs/polyglot-design.md`** documenting the design.
- **`src/Aspire.Hosting.AWS/README.md`** — new "Using a Node.js (TypeScript) AppHost" section with a feature/method-name table.
- **`playground/NodeAppHost/`** — a runnable Node.js AppHost sample exercising SDK config, DynamoDB Local, both Lambda models, and the API Gateway emulator. It consumes the locally built package via a local NuGet feed + a `pack-local.mjs` helper (versions as `100.0.<seconds-since-2020>` so the local build always wins). See its `README.md`.
- **Directory.Packages.props** — Aspire pin bumped to 13.4.6 (the version whose ATS codegen was validated) plus the transitive `Microsoft.Extensions.*` / `MessagePack` bumps that require.

## Testing

- **Unit tests** (`tests/Aspire.Hosting.AWS.UnitTests/`): `AtsExportTests` assert the expected `[AspireExport]`/`[AspireDto]` presence and the pinned capability ids; `LambdaPolyglotTests` cover project-path resolution and build behavior.
- **End-to-end**: `playground/NodeAppHost` runs under `aspire run` with all resources healthy. Invoking the API Gateway routes exercises the full path, including the executable Lambda reading/writing DynamoDB Local through the injected `AWS_ENDPOINT_URL_DYNAMODB` endpoint (a per-input invocation counter).

## Notes for reviewers

- The TypeScript-facing method names intentionally differ from the C# names where noted; there is no C# breaking change (C# method names are unchanged).
- `playground/NodeAppHost/packages/` is git-ignored (holds locally packed `.nupkg` files); only a placeholder `README.md` is committed to keep the folder present.


normj:
Build is failing do to a CVE issue with the CDK reference which has already been changed in PR https://github.com/aws/integrations-on-dotnet-aspire-for-aws/pull/219. I need to get the PR merged and then I can merge into this branch.

pjmagee:
exciting!

buildtools/ci.buildspec.yml
```diff
@@ -10,6 +10,7 @@ phases:
       - find / -type f -name 'global.json' -delete
       - curl -sSL https://dot.net/v1/dotnet-install.sh | bash /dev/stdin --channel 8.0
       - curl -sSL https://dot.net/v1/dotnet-install.sh | bash /dev/stdin --channel 10.0
+      - dotnet dev-certs https --trust || dotnet dev-certs https
```
normj, line 13: PR checks were failing because the dev cert was no longer trusted. Figured it didn't hurt to add it in the CI setup to trust the dev cert and then the PR check passed.

buildtools/ci.buildspec.yml
```diff
@@ -10,6 +10,7 @@ phases:
...
   build:
     commands:
       - dotnet test --verbosity normal integrations-on-dotnet-aspire-for-aws.sln --configuration Release --logger "console;verbosity=detailed" --logger trx --results-directory ./testresults
```
GarrettBeatty, line 16: are we sure the nodejs  playground tests are running in the ci? because i looked through the https://github.com/aws/integrations-on-dotnet-aspire-for-aws/actions/runs/30388990155/job/90375394693?pr=221 but didnt see them. and i see the command uses integrations aspir aws solution file but we didnt modify that at all for this pr so wondering how it picks up these new playground tests. can you double check?

normj: The `playground` area is not for tests in CI. That is more of adhoc area for all of the Aspire features to try out the user experience. The CI testing is geared towards ensuring the right attributes are in place and then relying on Aspire testing to work correctly for the attributes.

When I figure out how the deployment side of this can work which I think is going to be quite a bit more tricky I'm sure I'll need to revisit how this is testing and add unit/integ testing from the Node.js side.

GarrettBeatty (approved):
seems fine to me. will these code changes play well with https://github.com/aws/integrations-on-dotnet-aspire-for-aws/pull/224

src/Aspire.Hosting.AWS/Lambda/LambdaExtensions.cs
```diff
@@ -31,8 +31,68 @@ public static class LambdaExtensions
...
+    /// <param name="lambdaHandler">Lambda function handler</param>
+    /// <param name="options">Options for configuring the Lambda function.</param>
+    /// <returns></returns>
+    [AspireExport("addAWSLambdaFunction")]
+    internal static IResourceBuilder<LambdaProjectResource> AddAWSLambdaFunctionForPolyglot(this IDistributedApplicationBuilder builder, string name, string projectPath, string lambdaHandler, LambdaFunctionPolyglotOptions? options = null)
+    {
+        var functionOptions = new LambdaFunctionOptions();
+        if (!string.IsNullOrEmpty(options?.LogFormat))
+        {
+            functionOptions.LogFormat = Amazon.Lambda.LogFormat.FindValue(options.LogFormat);
```
dscpinheiro, line 53: This is the SDK's constant class, right? If yes, wouldn't it accept unknown values that are not valid?

(Same comment applies to the application log level)

src/Aspire.Hosting.AWS/Lambda/LambdaExtensions.cs
```diff
@@ -31,8 +31,68 @@ public static class LambdaExtensions
...
+                                        .ToArray();
+            if (projectFiles.Length != 1)
+            {
+                throw new DistributedApplicationException($"Path to Lambda project could not be determined. The directory '{projectPath}' must contain a single project file (*.csproj or *.fsproj).");
+            }
+
+            return projectFiles[0];
+        }
+
+        return projectPath;
```
dscpinheiro, line 91: If the method reaches here, doesn't it mean that the project path doesn't exist?

playground/NodeAppHost/pack-local.mjs
```diff
@@ -0,0 +1,43 @@
...
+import { rmSync, mkdirSync, readdirSync } from 'node:fs';
+
+const scriptDir = dirname(fileURLToPath(import.meta.url));
+const projectPath = join(scriptDir, '..', '..', 'src', 'Aspire.Hosting.AWS', 'Aspire.Hosting.AWS.csproj');
+const outputDir = join(scriptDir, 'packages');
+
+// aspire.config.json requests "Aspire.Hosting.AWS": "100.0.0", which NuGet treats as the range [100.0.0, ) and
+// resolves to the LOWEST matching version. So the local feed must contain exactly ONE 100.0.* package — the
+// freshly packed one. Clear the feed first so a previous (lower) dev version is not restored instead.
+rmSync(outputDir, { recursive: true, force: true });
```
dscpinheiro: Probably not a blocker but this deletes everything in the folder. Do we care about keeping `playground/NodeAppHost/packages/README.md`?

Directory.Packages.props
```diff
@@ -61,7 +61,7 @@
     <PackageVersion Include="xunit.runner.visualstudio" Version="3.1.5" />
     <PackageVersion Include="JsonSchema.Net" Version="7.4.0" />
     <PackageVersion Include="Moq" Version="4.20.72" />
-    <!-- Added explicity MessagePack reference to address NU1109 -->
-    <PackageVersion Include="MessagePack" Version="2.5.301" />
+    <!-- Added explicitly MessagePack reference to address NU1109 -->
```
dscpinheiro: I see this on my GitHub UI (extra carriage return?):

Image

src/Aspire.Hosting.AWS/Lambda/LambdaExtensions.cs
normj: The Aspire cross gen support doesn't work for our `ConstantClass` enums. I could maybe get to work if I added the Aspire attributes in the SDK but I don't want to do that. So the solution was to make a version of the DTO that used strings instead.

src/Aspire.Hosting.AWS/Lambda/LambdaExtensions.cs
normj: If the project path is a to a file not a directory then it would go there. Imagine project path set to `.\src\MyFunction\MyFunction.csproj`.

playground/NodeAppHost/pack-local.mjs
normj: Fixed

Directory.Packages.props
normj: Fixed

dscpinheiro (approved):

PR #230, opened 2026-08-18
Allow opting out of OTLP exporter wiring on AddAWSLambdaFunction

### LINKED ISSUE #228: Allow opting out of OTLP exporter wiring on `AddAWSLambdaFunction`
### Describe the feature

`AddAWSLambdaFunction` unconditionally calls the private `WithOpenTelemetry()` extension, which ends in Aspire's `WithOtlpExporter()`:

https://github.com/aws/integrations-on-dotnet-aspire-for-aws/blob/main/src/Aspire.Hosting.AWS/Lambda/LambdaExtensions.cs (see `resource.WithOpenTelemetry();` in `AddAWSLambdaFunction`, and the private `WithOpenTelemetry` method further down)

This injects `OTEL_EXPORTER_OTLP_ENDPOINT` into every Lambda function resource with no way to opt out: there is no option on the Lambda resource builder, nothing on `SQSEventSourceOptions`, and `WithOpenTelemetry` is private so it can't be overridden or skipped. A way to disable (or conditionally skip) this wiring would remove a hard-to-diagnose stall in integration testing scenarios.

### Use Case

We run integration tests against a real AppHost (LocalStack + two Lambda functions with `WithSQSEventSource`) via `Aspire.Hosting.Testing`'s `DistributedApplicationTestingBuilder`.

Under the testing builder no dashboard or OTLP collector runs, but the endpoint is still injected, pointing at an address nothing listens on. Each Lambda invoke then attempts to export telemetry to the dead endpoint, and the export/flush delay stalls the invoke and serializes the event pipeline — sub-second handler invocations become multi-second ones, making event-driven integration tests slow and timing-sensitive. The symptom (slow, serialized invokes) doesn't obviously point at OTel, so it's painful to diagnose the first time.

Our current workaround blanks the endpoint per Lambda resource in the test fixture:

```csharp
foreach (var lambdaName in new[] { "MyLambdaFunction", "MyOtherLambdaFunction" })
{
    var lambda = appHost.Resources.Single(r => r.Name == lambdaName);
    lambda.Annotations.Add(new EnvironmentCallbackAnnotation(env =>
        env["OTEL_EXPORTER_OTLP_ENDPOINT"] = string.Empty));
}
```

This works but is per-resource, relies on resource-name strings, and has to be rediscovered by every team that hits the stall.

### Proposed Solution

Either (or both) of:

1. **Make the wiring conditional**: skip `WithOtlpExporter()` when no OTLP collector/dashboard is configured for the application (e.g. under `DistributedApplicationTestingBuilder`, or whenever no dashboard is present), mirroring how a plain Aspire `AddProject` behaves in the same situation.
2. **Expose an opt-out**: e.g. a `DisableOpenTelemetry` option on `AddAWSLambdaFunction`, or make `WithOpenTelemetry()` public so callers can choose.

Option 1 fixes the default silently for everyone; option 2 is a smaller change that at least makes the workaround supported and discoverable. Happy to contribute a PR for either direction if maintainers indicate a preference.

### Other Information

The OTLP wiring is the right default when the AppHost runs with the dashboard — telemetry from Lambda invokes shows up in the dashboard and is genuinely useful in local dev. This request is only about the case where no collector exists to receive it, which today is silently pathological instead of a no-op.

### Acknowledgements

- [ ] I may be able to implement this feature request

- Aspire.Hosting.AWS 13.6.0
- Aspire.AppHost.Sdk / Aspire.Hosting.Testing 13.4.6
- LocalStack.Aspire.Hosting 13.1.0
- .NET 10 SDK, observed on macOS

### Targeted .NET Platform

.NET 10

macOS

dscpinheiro:
Resolves #228 by adding a `DisableOpenTelemetry` option that skips the `WithOpenTelemetry` call (default is false so existing behavior is unchanged).

## License
- [X] I confirm that this pull request can be released under the Apache 2 license

### DIFF
--- a/.autover/changes/cffc8829-e669-4515-92c4-176e543a8331.json
+++ b/.autover/changes/cffc8829-e669-4515-92c4-176e543a8331.json
@@ -1,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "Aspire.Hosting.AWS",
+      "Type": "Minor",
+      "ChangelogMessages": [
+        "Add DisableOpenTelemetry option to AddAWSLambdaFunction to opt out of automatic OTLP exporter wiring"
+      ]
+    }
+  ]
+}
--- a/src/Aspire.Hosting.AWS/Lambda/LambdaExtensions.cs
+++ b/src/Aspire.Hosting.AWS/Lambda/LambdaExtensions.cs
@@ -57,6 +57,8 @@ public static class LambdaExtensions
             functionOptions.ApplicationLogLevel = Amazon.Lambda.ApplicationLogLevel.FindValue(options.ApplicationLogLevel);
         }
 
+        functionOptions.DisableOpenTelemetry = options?.DisableOpenTelemetry ?? false;
+
         // suppressBuild: false so Aspire builds the project as part of `dotnet run`. Unlike the class library
         // wrapper project (which is pre-built by LambdaBeforeStartEventHandler), a polyglot project is not built ahead of time.
         return AddAWSLambdaFunctionCore(builder, name, lambdaHandler, new LambdaProjectMetadata(ResolvePolyglotProjectPath(builder, projectPath), suppressBuild: false), functionOptions);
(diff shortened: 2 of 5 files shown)


normj:
Looks the Aspire team merged in PR https://github.com/microsoft/aspire/pull/18918 a couple weeks back that now exposes the `WithProjectDefaults` as a public member. Once that goes out we can remove the Lambda custom logic for setting up OTEL and rely on that behavior. 

I would rather wait till that change is shipped instead of adding a custom property on our public interface.

dscpinheiro:
Sounds good, will close this then and reply to the original issue.

PR #231, opened 2026-08-18
Update to version Aspire 13.5

### LINKED ISSUE #228: Allow opting out of OTLP exporter wiring on `AddAWSLambdaFunction`

normj:
## Description
Update to Aspire 13.5.

The AddAWSLambdaFunction method now uses Aspire 13.5's new WithProjectDefaults method to configure LambdaProject subclasses with the same defaults that AddProject applies to the base Project type.

Switching to `WithProjectDefaults` address issue https://github.com/aws/integrations-on-dotnet-aspire-for-aws/issues/228 because we follow the same behavior of `AddProject` without attempting to mimic the behavior which has the discrepancy of how OTEL is setup.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -59,7 +59,7 @@
     <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="18.0.1" />
     <PackageVersion Include="xunit" Version="2.9.3" />
     <PackageVersion Include="xunit.runner.visualstudio" Version="3.1.5" />
-    <PackageVersion Include="JsonSchema.Net" Version="7.4.0" />
+    <PackageVersion Include="JsonSchema.Net" Version="9.4.0" />
     <PackageVersion Include="Moq" Version="4.20.72" />
     <!-- Added explicitly MessagePack reference to address NU1109 -->
     <PackageVersion Include="MessagePack" Version="2.5.302" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -2,48 +2,48 @@
   <PropertyGroup>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
     <CentralPackageTransitivePinningEnabled>true</CentralPackageTransitivePinningEnabled>
-    <AspireSdkVersion>13.4.6</AspireSdkVersion>
+    <AspireSdkVersion>13.5.0</AspireSdkVersion>
   </PropertyGroup>
   <ItemGroup>
-    <PackageVersion Include="Aspire.Hosting" Version="13.4.6" />
-    <PackageVersion Include="Aspire.Hosting.AppHost" Version="13.4.6" />
-    <PackageVersion Include="Aspire.Hosting.Testing" Version="13.4.6" />
-    <PackageVersion Include="Aspire.Hosting.Redis" Version="13.4.6" />
-    <PackageVersion Include="Aspire.Hosting.Valkey" Version="13.4.6" />
-    <PackageVersion Include="Aspire.Hosting.SqlServer" Version="13.4.6" />
-    <PackageVersion Include="Aspire.StackExchange.Redis" Version="13.4.6" />
-    <PackageVersion Include="Aspire.StackExchange.Redis.OutputCaching" Version="13.4.6" />
-    <PackageVersion Include="Microsoft.Extensions.ServiceDiscovery" Version="10.5.0" />
-    <PackageVersion Include="Microsoft.Extensions.ServiceDiscovery.Dns" Version="10.5.0" />
-    <PackageVersion Include="Microsoft.Extensions.Http" Version="10.0.8" />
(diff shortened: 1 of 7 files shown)


Directory.Packages.props
```diff
@@ -2,48 +2,48 @@
...
+    <PackageVersion Include="AWSSDK.BedrockAgentCore" Version="4.0.105.1" />
+    <PackageVersion Include="AWSSDK.CloudFormation" Version="4.0.102.2" />
+    <PackageVersion Include="AWSSDK.Core" Version="4.0.101.1" />
+    <PackageVersion Include="AWSSDK.DynamoDBv2" Version="4.0.103.3" />
+    <PackageVersion Include="AWSSDK.Lambda" Version="4.0.105.2" />
+    <PackageVersion Include="AWSSDK.S3" Version="4.0.102.2" />
+    <PackageVersion Include="AWSSDK.SecurityToken" Version="4.0.100.9" />
+    <PackageVersion Include="AWSSDK.SimpleNotificationService" Version="4.0.100.9" />
+    <PackageVersion Include="AWSSDK.SQS" Version="4.0.100.9" />
     <PackageVersion Include="AWSSDK.Signin" Version="4.0.2" />
```
dscpinheiro: Is not updating `AWSSDK.Signin` only intentional?

dscpinheiro (approved):

Directory.Packages.props
normj: No that was not intentional. Pushed new commit with updating AWSSDK.Signin

normj:
Putting this in draft mode because the switch to `WithProjectDefaults` is causing a problem for class library based projects when run outside of the debugger. Need to investigate what we can do.

GarrettBeatty (approved):

